PR #446, opened 2014-08-29
persian sura name + supportsRtl

amirhakh:
In the name of Allah

add arabic sura_name to "values-fa"
enable supportsRtl
add arrabic layout to layout-rtl


amirhakh:
Salam

Dose not matter compatibility issu, this code (xml) not affect on prevuse android. This will seen on andrroid 4.2+ .If you add this line to AndroidManifest, get warning about that not affect on old android.
Only you need android sdk 4.2+ to compile this code and  if android version above that, load it property.


فی امان الله

ahmedre:
wa3laikum alsalam,
i removed `supportsRTL` because it breaks the arabic ui on the device. for now, due to the fact that we have to support pre-4.2 devices, we will stick with `-ar` or `-rtl` folders instead of using `supportsRTL` (having both causes post-4.2 to break).
walsalam 3alaikum.

ahmedre:
ok, i just tried this locally, and seems like `layout-rtl` isn't quite respected in farsi for some reason - so can you base something off of master and just copy the `-ar` folders to `-fa` and test it and see if it works fine? for the ordering of the items in the first screen, we can change the code to correctly reverse the pages as well.

amirhakh:
salam

I try to fix rtl support for this project in Nooruz (نوروز) and comit you my result. <In sha allah>
tank you &
اجرکم عند الله

ozbek:
@amirhakh: you may probably want to close this PR now.

amirhakh:
salam

yes

i was make new pull request

PR #473, opened 2015-01-09
Russian translation update

Valey:
Russian translation update:

    update strings.xml
    add sura_name.xml
    add readers.xml


Valey:
Russian translation update:
 - update strings.xml
 - add sura_name.xml
 - add readers.xml

app/src/main/res/values-ru/sura_names.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string-array name="sura_names">
+    <item>аль-Фатиха (Открывающая Книгу)</item>
```
ozbek: Where did you get the Sura Names from?
Normally, prefixes are capitalized, but your version are all with lower case.

Valey: I verified with 2 translation. In the first one all letters was capitalized, but in the second one prefixes was lowercase. I'll check it again

ozbek: Thanks. 
Just to be clear, I wanted to say _the first letters of prefixes are capitalized_ (not whole prefix); i.e. Аль-Фатиха (not АЛЬ-Фатиха).

Valey:
Yes, I understood. My imam said that it doesn't matter, so I changed prefixes to capitalized.

ahmedre:
jazakAllah khairan!

PR #479, opened 2015-02-23
L

shofwan:


ozbek:
Ouch, why?!

ahmedre:
jazakAllah khairan, but we don't want to merge to master yet until we're ready to release insha'Allah, especially because we need to figure out where to branch a "3.x and below" branch from.

PR #480, opened 2015-02-23
Fix 'Too Long Log Tags' lint error

ozbek:
Lint is complaining with

 _The logging tag can be at most 23 characters, was NN.
  Log tags are only allowed to be at most 23 tag characters long._

error message.


ahmedre:
jazakAllah khairan!

PR #481, opened 2015-03-01
Sync external SD card pref title with that of in StorageUtils

ozbek:
This is how it looks now:

screenshot_2015-03-01-21-47-14

It feels a little confusing with external storage number (1) and the free space amount (59933) being shown side-by-side. Probably, free space amount needs to be shown inside parenthesis or otherwise made it visually distinguishable.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
+++ b/app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
@@ -37,8 +37,8 @@ public class StorageUtils {
 
     final File[] mountPoints = ContextCompat.getExternalFilesDirs(context, null);
     if (mountPoints != null && mountPoints.length > 1) {
-      for (int i = 0; i < mountPoints.length; i++) {
-        mounts.add(mountPoints[i].getAbsolutePath());
+      for (File mountPoint : mountPoints) {
+        mounts.add(mountPoint.getAbsolutePath());
       }
     } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT) {
       mounts = readMountsFile();
@@ -48,9 +48,8 @@ public class StorageUtils {
       if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2) {
         Collection<String> vold = readVoldsFile();
 
-        List<String> toRemove = new ArrayList<String>();
-        for (Iterator<String> iter = mounts.iterator(); iter.hasNext(); ) {
-          String mount = iter.next();
+        List<String> toRemove = new ArrayList<>();
+        for (String mount : mounts) {
           if (!vold.contains(mount)) {
             toRemove.add(mount);
           }
@@ -71,7 +70,7 @@ public class StorageUtils {
 
   private static List<Storage> buildMountsList(Context context,
                                                Collection<String> mounts) {
-    List<Storage> list = new ArrayList<Storage>(mounts.size());
+    List<Storage> list = new ArrayList<>(mounts.size());
 
     int externalSdcardsCount = 0;
     if (mounts.size() > 0) {
@@ -108,7 +107,7 @@ public class StorageUtils {
   private static Collection<String> readMountsFile() {
     String sdcardPath = Environment
         .getExternalStorageDirectory().getAbsolutePath();
-    List<String> mounts = new ArrayList<String>();
+    List<String> mounts = new ArrayList<>();
     mounts.add(sdcardPath);
 
     Log.d(TAG, "reading mounts file begin");
@@ -146,7 +145,7 @@ public class StorageUtils {
    * reads volume manager daemon file for auto-mounted storage
    * read more about it here: http://vold.sourceforge.net/
    *
-   * @return
+   * @return Mount points from `vold.fstab` configuration file
    */
   private static Set<String> readVoldsFile() {
     Set<String> volds = new HashSet<>();
@@ -192,7 +191,6 @@ public class StorageUtils {
     private String label;
     private String mountPoint;
     private int freeSpace;
-//    private int totalSpace;
 
     public Storage(String label, String mountPoint) {
       this.label = label;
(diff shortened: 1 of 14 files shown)


ahmedre:
i am confused about how your fix fixes this though - you are converting "External SD Card" to "External SD Card 1" in this patch - before your patch, i take it that we used to just display "External SD Card," right?

ozbek:
Yes, that's right. If you see the `StorageUtils`, that's how the preference is titled: _External SD Card 1_, _External SD Card 2_, etc.

Are there user devices with more than one SD card slots? Maybe it should be synced other way round and card numbers should be removed from `StorageUtils`?

In fact, I am not sure why `QuranSettingsFragment` is not using `StorageUtils` at this point and trying to discover mount points directly (probably should in order to avoid DRY).

ahmedre:
ah - i see - yes we're adding the number in `StorageUtils` indeed - so a few things:
1. i agree - any of that storage stuff should live in `StorageUtils`.
2. `StorageUtils` can be cleaned up - it has a bunch of 2.3 and below code that can now be removed.
3. ideally, formatting like this should live in strings - i.e `External SD Card %1$d (%2$d)` or something similar.

ozbek:
This is a quick clean-up and implementation of the points you have mentioned. Only tested on Note 4 and Nexus 5, no exhausted  tests are done.

Please let me know your opinion on the changes. I see at least other strings/locales should be fixed.

app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
```diff
@@ -256,11 +233,11 @@ public int getFreeSpace() {
       return freeSpace;
     }
 
-    /**
-     * @return total size in Megabytes
-     */
-    public int getTotalSpace() {
-      return totalSpace;
-    }
+//    /**
```
ahmedre: remove it instead

app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
```diff
@@ -232,12 +202,19 @@ public Storage(String label, String mountPoint) {
...
-      long bytesAvailable = (long) stat.getAvailableBlocks() *
-          (long) stat.getBlockSize();
+//      long totalBytes;
+      long bytesAvailable;
+      if (Build.VERSION.SDK_INT > Build.VERSION_CODES.JELLY_BEAN_MR1) {
+//        totalBytes = stat.getBlockCountLong() * stat.getBlockSizeLong();
+        bytesAvailable = stat.getAvailableBlocksLong() * stat.getBlockSizeLong();
+      } else {
+        //noinspection deprecation
+//        totalBytes = (long) stat.getBlockCount() * (long) stat.getBlockSize();
```
ahmedre: remove the stuff you commented out, we can re-add if we need it later

app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
```diff
@@ -222,7 +192,7 @@ private static boolean isExternalStorageEmulatedHoneycomb() {
     private String label;
     private String mountPoint;
     private int freeSpace;
-    private int totalSpace;
+//    private int totalSpace;
```
ahmedre: remove

app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
```diff
@@ -68,35 +78,11 @@
       String firstItem = mounts.iterator().next();
 
       // Follow Android SDCards naming conventions
-      if (Build.VERSION.SDK_INT < Build.VERSION_CODES.GINGERBREAD) {
-        list.add(new Storage(
-            context.getString(R.string.prefs_sdcard_auto),
```
ahmedre: this isn't used anymore, can we remove the string from the strings files? speaking of which, we need a utility to make this easier, it's a hassle right now :(

app/src/main/res/values/strings.xml
```diff
@@ -168,7 +168,7 @@
     <string name="prefs_app_location_title">Quran Data Directory</string>
     <string name="prefs_app_location_summary">Choose where to store Quran files.</string>
     <string name="prefs_sdcard_internal">Internal Storage</string>
-    <string name="prefs_sdcard_external">External SD Card</string>
+    <string name="prefs_sdcard_external">External SD Card %1$d</string>
```
ahmedre: either need to change it for other languages as well or remove the string from all non-English languages so people can add them again.

app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
```diff
@@ -33,34 +33,44 @@
...
+      if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2) {
         Collection<String> vold = readVoldsFile();
 
         List<String> toRemove = new ArrayList<String>();
-        for (Iterator<String> iter = mounts.iterator(); iter.hasNext(); ){
-            String mount = iter.next();
-            if (!vold.contains(mount)){
-                toRemove.add(mount);
-            }
+        for (Iterator<String> iter = mounts.iterator(); iter.hasNext(); ) {
```
ahmedre: `foreach (String mount : mounts) {`

ahmedre:
just some minor suggestions, otherwise looks good to me. the note 4 has an external sdcard i guess, so that's the better use case. i don't have a 4.x+ phone that has an external sdcard so it's also difficult for me to test.

ozbek:
Removed the commented out code and fixed other warning Android Studio gave in `StorageUtils`. 

Added number reference to all `prefs_sdcard_external` translations except for Kurdish (values-ku). This one has mixture of Latin and Arabic text, so I could not figure out where to correctly put the wildcard. But I neither remove the translation (as you have suggested). The current state should be good without any runtime issues. The only thing is number reference will go void at `getString()` for Kurdish.

ahmedre:
jazakAllah khairan - i manually merged it.

Issue #474, opened 2015-01-11
Russian translation mistakes


Somaiagabr:
Ya-Sin missing from the first ayat of Ya-Sin Russian translation Arabic text instead of Ya-Sin there is Bismilah.

ozbek:
If you look a little carefully, you will recognize that Ya-Sin is there right after the basmala. 

In fact that is the case with the other Surahs, too: basmala is prefixed to the the first ayah. 
I always thought that was something done intentionally...

mhmgad:
Salam Alikom

@ozbek I don't  contribute to the app, but many scholars consider Basmala part of the first Ayah for all the Surahs except Al-tawbah.

It's almost agreed that for Fat-ha (1st surah) basmala is the 1st ayah. Other Surahs you may find different scholars opinions but It makes no different.

ozbek:
Wa alaikum assalam

@mhmgad we are talking about same thing, but in different words :)

@Somaiagabr can you confirm that Ya-Sin is actually not missing?

hakim89:
Salam Alykom, 
Can we please wrap values-ru xml tags in <![CDATA[] tags becuase build failed becuase of parse error in that file, i did wrap them up in my project but this messes up the russian text becuase you dont need escape chars anymore

ozbek:
Wa alaikum assalam 
Can you please post the build errors?

hakim89:
[Fatal Error] :118:5: The element type "item" must be terminated by the matching end-tag "</item>".
/Users/hakim/Documents/MyProjects/quran_android/app/src/main/res/values-ru/sura_names.xml:117:4: Error: The element type "item" must be terminated by the matching end-tag "</item>".
:app:mergeDebugResources FAILED

FAILURE: Build failed with an exception.

ozbek:
Thanks, @hakim89. Fixed, let's wait for project maintainers to merge...

ahmedre:
@ozbek can we close this?

ozbek:
Yes, as far as I can see all reported problems here are already fixed.

ahmedre:
jazakAllah khairan

PR #482, opened 2015-03-07
Add missing closing tag

ozbek:
This is to fix the issue reported by @hakim89 in https://github.com/quran/quran_android/issues/474


ahmedre:
ah, on master, i think i fixed it on l sometime back after someone sent a patch for this - jazakAllah khairan for fixing this!

PR #483, opened 2015-03-07
Use getExternalStorageDirectory() for internal SD card path

ozbek:
This is needed for correctly compairing the user selected value with that of `mInternalSdcardLocation` in QuranSettingsFragment (line 61):

    mInternalSdcardLocation =
        Environment.getExternalStorageDirectory().getAbsolutePath();

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
+++ b/app/src/main/java/com/quran/labs/androidquran/util/StorageUtils.java
@@ -11,12 +11,8 @@ import com.quran.labs.androidquran.R;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Scanner;
-import java.util.Set;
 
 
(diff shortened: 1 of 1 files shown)


ozbek:
Please see the follow-up CL. Hopefully, I understood you correctly :)

ahmedre:
yep, you understood right - one small thing though, i think volds should remain a set due to the way we're checking if items are in it or not. i'll merge this patch and then can you re-fix that? jazakAllah khairan.

ahmedre:
btw, on an unrelated note - planning on releasing this really soon - maybe even today. i know it's not "finished" (especially the new audio manager, etc) - but:

- we haven't released anything since July of last year (almost a year) - sometimes, no updates for a long time make people think an app is dead
- releasing lets us figure out bugs and fix them (before Ramadan, which is the busy time)
- the "todo" list is endless, and so not releasing due to "we need to add things," means we may never release :)

ozbek:
Yes, I also agree that this time round the updates got delayed for too long.

Regarding the "contributors", I will wait for the redo...

Issue #491, opened 2015-03-10
What's the purpose of preference classes?


ozbek:
At the moment there are 4 preference files under [`app/src/main/java/com/quran/labs/androidquran/ui/preference`](https://github.com/quran/quran_android/tree/master/app/src/main/java/com/quran/labs/androidquran/ui/preference). Except for `SeekBarPreference.java`, the other 3 are doing only one thing: **change the color of the dialog title**.

However, it appears to me that such change was required sometime earlier but is not required anymore. For instance, I compared runtime results for `QuranListPreference` and `ListPreference` and they look identical on a device.

Maybe those classes are not needed anymore?

ahmedre:
i don't remember on the top of my head, but you maybe right, we may not need this anymore. i will test it and get back to you insha'Allah.

ozbek:
Please see the discussion in PR https://github.com/quran/quran_android/pull/492 for details.

PR #492, opened 2015-03-10
Remove unnecessary custom preference implementations

ozbek:
Fixes https://github.com/quran/quran_android/issues/491


ahmedre:
they actually are different (it makes the item title font colors white, which looks a lot more readable imho) - here are some screenshots:

with Quran preference implementation

without Quran preference implementation

ozbek:
I see, thanks for verifying.

Probably, that's device/emulator implementation dependent: I don't see any color difference on Galaxy S5 and Nexus 5. Same bright white color both with and without custom preferences.

Issue #494, opened 2015-03-14
Bottom Play Controls Next Ayah and Previous Ayah are working the opposite


malzantot:
Salam Alikum,

I noticed that there is a problem with the audio controls that appear in the bottom bar while playing audio.
**The next and previous buttons are working as opposite**. Currently the forward button (pointing to the right) moves to previous Aya,  while the rewind button (pointing to the left) moves to next Ayah. I would expect that right button (forward) moves to next as it is pointing to the same direction as the play button. Moreover, **they are inconsistent with the play controls that appear in the notification bar**.

To reproduce:

1. Start playing audio of any Surah.
3. Try the forward and rewind buttons in the bottom bar and compare its effect against the next and previous buttons in the notification bar.

<img src="https://cloud.githubusercontent.com/assets/1429452/6653449/a85c4b86-ca50-11e4-9055-a165a6b91f64.png" style="height: 100" />

PR #495, opened 2015-03-14
Interchanging functionality of Skip and Rewind buttons in the Bottom Audio Status Bar

malzantot:
Salam Alikum ,

Here I am fixing the issue#494 https://github.com/quran/quran_android/issues/494 regarding the *skip* and *rewind* buttons in the bottom AudioStatusBar.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/widgets/AudioStatusBar.java
+++ b/app/src/main/java/com/quran/labs/androidquran/widgets/AudioStatusBar.java
@@ -369,10 +369,10 @@ public class AudioStatusBar extends LinearLayout {
                   mAudioBarListener.onPausePressed();
                   break;
                case R.drawable.ic_next:
-                  mAudioBarListener.onPreviousPressed();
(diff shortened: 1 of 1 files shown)


ahmedre:
salam 3alaikum,
i think we should keep `AudioStatusBar` as it is - the order was put that way with the intention of matching the page structure - right goes to a previous page (previous ayah also, since right to left), and left goes to the next one.

that being said, i'd fix the inconsistency by swapping the order in the notifications - most people probably don't know about the notification controls, so changing those is bound to bring less complaints than changing the buttons at the bottom.

can you submit a new patch? you want to look at `AudioService`.
jazakAllah khairan.
walsalam 3alaikum.

ahmedre:
btw, forgot to say jazakAllah khair, jazakAllah khairan for pointing this out (and jazakumAllah khairan for the patch!)

malzantot:
Salam Alikum,

Jazkum Allah Khayran for your reply.. I have created another pull request implementing your suggestion.

PR #496, opened 2015-03-15
Swapping the actions of skip, and rewind controls in the notification bar

malzantot:
Salam Alikum ,

Here I am fixing the issue#494 https://github.com/quran/quran_android/issues/494 by swapping the actions of the notification bar skip and rewind buttons as suggested.


ahmedre:
jazakAllah khairan

PR #501, opened 2015-03-19
Put free space information below the label

ozbek:
This was somewhat mentioned in PR https://github.com/quran/quran_android/pull/481:
original implementation put storage label, number and free space in one line; that feels and looks confusing. This is an attempt to give the list a more decent look. 
Following is how it looks now:

device-2015-03-19-205313

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/preference/DataListPreference.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/preference/DataListPreference.java
@@ -1,0 +1,127 @@
+package com.quran.labs.androidquran.ui.preference;
+
+import android.annotation.TargetApi;
+import android.app.Activity;
+import android.app.AlertDialog;
+import android.content.Context;
+import android.graphics.Color;
+import android.os.Build;
+import android.preference.ListPreference;
+import android.support.annotation.NonNull;
+import android.text.TextUtils;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.CheckedTextView;
+import android.widget.ListAdapter;
+import android.widget.TextView;
+
+import com.quran.labs.androidquran.R;
+import com.quran.labs.androidquran.util.QuranSettings;
+import com.quran.labs.androidquran.util.StorageUtils;
+
+import java.util.List;
+
+public class DataListPreference extends ListPreference {
+  private CharSequence[] mFreeSpaceAmounts;
+
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+  public DataListPreference(Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes) {
+    super(context, attrs, defStyleAttr, defStyleRes);
+  }
+
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+  public DataListPreference(Context context, AttributeSet attrs, int defStyleAttr) {
+    super(context, attrs, defStyleAttr);
+  }
+
+  public DataListPreference(Context context, AttributeSet attrs) {
+    super(context, attrs);
+  }
+
+  public DataListPreference(Context context) {
+    super(context);
+  }
+
+  public void setLabelsAndSummaries(Context context, int appSize,
+                                    List<StorageUtils.Storage> storageList) {
+    String summary = context.getString(R.string.prefs_app_location_summary) + "\n"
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/preference/DataListPreference.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public View getView(int position, View convertView, ViewGroup parent) {
+      if (convertView == null) {
+        LayoutInflater inflater = ((Activity) getContext()).getLayoutInflater();
+        convertView = inflater.inflate(R.layout.data_storage_location_item, parent, false);
+      }
+
+      TextView label = (TextView) convertView.findViewById(R.id.storage_label);
```
ahmedre: in another patch, cache these - i.e.:

```java
static class ViewHolder {
  public TextView title;
  public TextView subtitle;
}

// in getView():
ViewHolder holder;
if (convertView == null) {
  // inflate
  holder = new ViewHolder();
  holder.title = (TextView) findViewById(R.id.available_free_space);
  // .. and so on
  convertView.setTag(holder);
}
holder = (ViewHolder)convertView.getTag();
holder.title.setText("whatever");
```

app/src/main/java/com/quran/labs/androidquran/ui/preference/DataListPreference.java
```diff
@@ -0,0 +1,127 @@
...
+import com.quran.labs.androidquran.util.QuranSettings;
+import com.quran.labs.androidquran.util.StorageUtils;
+
+import java.util.List;
+
+public class DataListPreference extends ListPreference {
+  private CharSequence[] mFreeSpaceAmounts;
+
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+  public DataListPreference(Context context, AttributeSet attrs, int defStyleAttr, int defStyleRes) {
```
ahmedre: in another patch:
1. remove lollipop constructor (they shouldn't be called in these cases anyway)
2. only one of these constructors is lollipop only (the first one).
3. proxy each constructor to the other and run all the code in one... this makes it easier if we need to do something in common for all these constructors.

i.e.

```java
public DataListPreference(Context context) {
   this(context, null);
}

public DataListPreference(Context context, AttributeSet attrs) {
   this(context, attrs, 0);
}

public DataListPreference(Context context, AttributeSet attrs, int defStyleAttr) {
   super(context, attrs, defStyleAttr);
}
```

app/src/main/res/layout/data_storage_location_item.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ahmedre: RelativeLayout

app/src/main/res/layout/data_storage_location_item.xml
```diff
@@ -0,0 +1,45 @@
...
+              android:layout_width="match_parent"
+              android:layout_height="wrap_content"
+              android:orientation="horizontal"
+              android:paddingLeft="16dp"
+              android:paddingRight="16dp"
+              android:minHeight="?android:attr/listPreferredItemHeight">
+
+  <LinearLayout
```
ahmedre: kill weights, make it `alignParentStart` and `alignParentLeft`, and make it `toLeftOf` the box, which is `alignParentRight` and `alignParentEnd`.

ahmedre:
added some comments - i'll commit for now, but try to clean up if you get a chance. also, imho, mb might not be the best measurement unit anymore. we should probably show the unit depending on how much free space there actually is.

ahmedre:
jazakAllah khair.

ozbek:
Thanks for the comments, I will get back to them ASAP.

app/src/main/java/com/quran/labs/androidquran/ui/preference/DataListPreference.java
ozbek: Regarding the proxying (point 3.), would not this cause a failure on older platforms where the super has no three-argument constructor?

ozbek: Just checked with Nexus S emulator API 18, proxying is throwing `NoSuchMethodException`. 

I think we still need to keep the constructors.

ozbek:
I kept the constructors, other than that I tried to implement all of your suggestions. 
Jazak Allah khair.

PR #504, opened 2015-03-22
Fixing dropdown width for Audio panel

aessam:


ahmedre:
manually merged, jazakAllah khairan.

PR #509, opened 2015-03-24
Clean-up DataListPreference

ozbek:
- Added some comments.
 - Extended from `QuranListPreference` to reduce code-duplication.
 - Added `ViewHolder` sub-class for caching.
 - Replaced `LinearLayout` with `RelativeLayout` for better layout management.

Please see https://github.com/quran/quran_android/pull/501 for related discussion.


ahmedre:
looks good, jazakAllah khair. regarding the constructor thing, 2 of them only are low api versions (the other two are lollipop+) - so if we kept only the 2 with lower api versions and chained them, there should be no problem.

PR #515, opened 2015-04-06
Update code style guide

ozbek:
Hopefully, this will put all of contributors to the same track and avoids unnecessary code style changes in the future CLs.


ahmedre:
jazakAllah khairan!

PR #541, opened 2015-06-16
Changes the Arabic transliteration for sura names. 

muhsinali:
Endings were inconsistent and changes were made to reflect Darussalam's English transliterations & quran.com titles.

### DIFF
--- a/app/src/main/res/values/sura_names.xml
+++ b/app/src/main/res/values/sura_names.xml
@@ -1,40 +1,40 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
   <string-array name="sura_names">
-    <item>Al-Fatiha</item>
-    <item>Al-Baqara</item>
-    <item>Aal-E-Imran</item>
-    <item>An-Nisa</item>
-    <item>Al-Maeda</item>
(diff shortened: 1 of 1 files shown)


ahmedre:
jazakAllah khairan!

ahmedre:
@muhsinali one minor thing - some of these match what we have on quran.com, but i don't agree with them - for example, adding an ' before Al-'Isra - i think Al-Isra' is more correct. i found this (http://www.noblequran.com/translation/) which is basically almost what you have changed, but with a few of the apostrophes changed.

do you think those may end up being a little better (if we can make those few changes?) jazakAllah khairan!

muhsinali:
@ahmedre not a problem insha'allah I can make those changes. Do you want all the titles to match http://www.noblequran.com/translation/ ? or specifically al-Isra'? Just let me know and I'll update.

ahmedre:
ideally all of them if it's not too much trouble - i imagine the spellings should mostly match, it's just the extra ' that will likely change. should likely change them on quran.com as well insha'Allah. jazakAllah khairan!

muhsinali:
Not a problem at all insha'allah

PR #544, opened 2015-06-16
Updated changes to transliteration surah names.

muhsinali:
These are updated directly from titles at: http://www.noblequran.com/translation/

If individual tweaks need to be made let me know, however i pulled the names directly from the site itself.

### DIFF
--- a/app/src/main/res/values/sura_names.xml
+++ b/app/src/main/res/values/sura_names.xml
@@ -3,25 +3,25 @@
   <string-array name="sura_names">
     <item>Al-Fatihah</item>
     <item>Al-Baqarah</item>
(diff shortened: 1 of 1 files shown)


ahmedre:
sorry, i didn't look well enough yesterday, some of our spellings are a lot better than those on the site.
let's keep the ones from the previous patch, and from this one, let's only keep Isra' and Anbiya' from noblequran.

i'll close this one and i'll fix them insha'Allah - jazakAllah khair and really sorry for making you do extra work :( may Allah put it in mizan 7asanatak.

ahmedre:
https://github.com/quran/quran_android/commit/e1408527c3e0031a2b040f0d74b9c6ba25eefeac

PR #608, opened 2016-01-07
Update playback notification

ozbek:
Now it is possible to control the playback from the lock screen without expanding the notification.

Icons are re-created using [Android Asset Studio](https://romannurik.github.io/AndroidAssetStudio/icons-generic.html):
 - Size: 44 dip
 - Padding: 2 dip
 - Color: white (100%)


ahmedre:
masha'Allah man, i did this in haramain some time back and wanted to do it in Quran as well, but you beat me to it, may Allah reward you!

just some minor quibbles:

screen shot 2016-01-07 at 10 19 38 pm

this looks a bit jarring now compared to what it was (mostly due to the big icons). furthermore, the icon adjustment in size makes them look a bit too big here:

screen shot 2016-01-07 at 10 20 05 pm

i am not sure if it's possible, but can we see if we can you change it so that we keep the lock screen controls without changing the playback style of the notification? otherwise, if not (my guess is it's not easy to split these), we'll need a nicer icon in these cases rather than the app icon.

i also think this looks awesome but the icons for controls are too big:
screenshot-2016-01-07_22 15 58 17

i think we can revert the icon size changes - here's an example of haramain running with the original sizes that were in quran:

screen shot 2016-01-07 at 10 20 56 pm


and the notification:
screen shot 2016-01-07 at 10 18 10 pm

ahmedre:
so in summary - can you undo the resizing of the playback controls, and see what we can do about the icon in the notification - can we hide it? if not, if i can try to see if @Somaiagabr can come up with something that we can put there, and if not, we have higher resolution versions of the icon that we could use so it doesn't look grainy, etc.

ozbek:
> this looks a bit jarring now compared to what it was

Jarring look is due to icon size difference: loop and settings icons are too small(!). I noticed that, but did not go about changing them also, because this PR was meant to enable controls in lock screen.

I think, that's a matter of taste: I like big controls :)
Anyhow, as a Russian saying goes _"no disputing about tastes"_, will revert the icons back.

ahmedre:
i'll give you that it does look good in the lock screen, though it does take up more space :) looks like Google Play Music does [a bit of both](http://www.androidpolice.com/2015/12/13/the-latest-version-of-google-play-music-adds-thumbs-up-and-thumbs-down-to-the-notification-controls-apk-download/) - big icons for the "most likely action," and smaller icons for the others.

ozbek:
Done :)

ozbek:
Play Music is, most probably, using separate icons for notification and in-app controls.

ahmedre:
i am ok if we do the same - still think they'll be a bit too big on the bottom bar within the app though.

btw, i have a script that automatically grabs the images from google's github and places them in the right place - i'll check it in utils - it can save you time in the future insha'Allah :)

suggestions about the Quran icon in the notification shade? accepting this for now, maybe start an issue to discuss how to improve it?

jazakAllah khairan :)

PR #610, opened 2016-01-10
Implement nicer looking notifications.

ahmedre:
This patch draws a custom "album art" image for use in the media style.
It looks best on apis 21+, where the notification background color can
be changed, but looks fine pre-21 as well. Furthermore, since there's a
possibility that generating this image throws OOM, fall back to the
previous, non-MediaStyle notification if we can't generate the image.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
+++ b/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
@@ -43,9 +43,12 @@ import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.res.AssetFileDescriptor;
+import android.content.res.Resources;
 import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.Canvas;
+import android.graphics.drawable.ColorDrawable;
 import android.media.AudioManager;
 import android.media.MediaPlayer;
 import android.media.MediaPlayer.OnCompletionListener;
@@ -199,6 +202,7 @@ public class AudioService extends Service implements OnCompletionListener,
 
   private int mGaplessSura = 0;
   private int mNotificationColor;
+  private Bitmap mNotificationIcon;
   private SparseIntArray mGaplessSuraData = null;
   private AsyncTask<Integer, Void, SparseIntArray> mTimingTask = null;
 
@@ -282,9 +286,9 @@ public class AudioService extends Service implements OnCompletionListener,
     mMediaSession = new MediaSessionCompat(appContext, "QuranMediaSession");
 
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-      mNotificationColor = appContext.getResources().getColor(R.color.notification_color, null);
+      mNotificationColor = getResources().getColor(R.color.audio_notification_color, null);
     } else {
-      mNotificationColor = appContext.getResources().getColor(R.color.notification_color);
+      mNotificationColor = getResources().getColor(R.color.audio_notification_color);
     }
   }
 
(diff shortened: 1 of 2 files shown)


ahmedre:
cc @ozbek 

from 5.x emulator
screenshot-2016-01-09_22 12 19 797

if no memory to make the image on 5.x:
screenshot-2016-01-09_22 13 04 956

on 4.x:
screenshot-2016-01-09_22 12 12 726

if no memory on 4.x:
screenshot-2016-01-09_22 12 56 469

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
```diff
@@ -1081,14 +1107,17 @@ private void setUpAsForeground() {
...
-          .addAction(R.drawable.ic_previous, getString(R.string.previous), previousIntent)
-          .addAction(R.drawable.ic_pause, getString(R.string.pause), pauseIntent)
-          .addAction(R.drawable.ic_next, getString(R.string.next), nextIntent)
-          .setLargeIcon(icon)
-          .setShowWhen(false)
-          .setStyle(new NotificationCompat.MediaStyle()
-              .setShowActionsInCompactView(new int[]{0, 1, 2})
-              .setMediaSession(mMediaSession.getSessionToken()));
+          .addAction(R.drawable.ic_previous,
+              emptyTitles ? "" : getString(R.string.previous), previousIntent)
```
ozbek: What's the logic behind this: `emptyTitles ? "" : getString(R.string.previous)`? Why does playback control button's titles should depend on large icon?

ahmedre: see the screenshots on https://github.com/quran/quran_android/pull/610 - basically, when there is no large notification icon, if we still set `MediaStyle`, it will make the small icon stretch to fit and looks really bad - consequently, that logic is to behave the old way (buttons without titles) when we can't load the large image (and we don't `setStyle` and we also don't `setLargeIcon` in those cases).

ozbek: Sorry, I still don't get it. How does the empty title affect the the button size?

ahmedre: it doesn't effect the button size - it just looks really bad without it - you can try it - try on a pre-5.x emulator, remove the check and `""` fallback, and set the icon to null. then try it with it on.

ozbek:
I see. So, you simply do not want the text?

Without text, the buttons do not move to the center though and they look misaligned.

screenshot_2016-01-10-15-36-22

screenshot_2016-01-10-15-40-06

ahmedre:
that was the previous behavior - but if it bothers you, i'll accept a pull request to always have the text :)

ahmedre:
btw, generally agree with you - i forgot to mention that there's a second reason for this:

screenshot-2016-01-10_23 19 38 845

on smaller screens (this is a galaxy s2 emulator, so 480x800), previous is cut off.

note that this screenshot shows another bug, which is that we have mixed arabic/english - since the ui is arabic, the notification text should also be arabic, but for some reason it's english here...

ozbek:
Previous is truncated at the end, that is expected behavior (not a bug: the text is too long). 

Mixture of Arabic/English is inevitable: Arabic is not natively supported. I have seen platform get confused when locale configurations are managed internally in the app. 

Also, I didn't check yet: have previous/pause/next been translated into Arabic? I added those words very recently.

ahmedre:
ah, excellent point - that would be why - i don't think we translated them yet. i am cool if we remove this for code simplicity (also, this is arguably an edge case anyway, since we hope that in most cases, insha'Allah people will get the normal view).

PR #612, opened 2016-01-10
Add setWhen(0)

ozbek:
On older platforms, the time is shown even though `setShowWhen(false)` is set.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
+++ b/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
@@ -1111,10 +1111,11 @@ public class AudioService extends Service implements OnCompletionListener,
               emptyTitles ? "" : getString(R.string.previous), previousIntent)
           .addAction(R.drawable.ic_pause, emptyTitles ? "" : getString(R.string.pause), pauseIntent)
           .addAction(R.drawable.ic_next, emptyTitles ? "" : getString(R.string.next), nextIntent)
-          .setShowWhen(false);
+          .setShowWhen(false)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
```diff
@@ -1111,10 +1111,11 @@ private void setUpAsForeground() {
               emptyTitles ? "" : getString(R.string.previous), previousIntent)
           .addAction(R.drawable.ic_pause, emptyTitles ? "" : getString(R.string.pause), pauseIntent)
           .addAction(R.drawable.ic_next, emptyTitles ? "" : getString(R.string.next), nextIntent)
-          .setShowWhen(false);
+          .setShowWhen(false)
```
ahmedre: yeah looks like the real call was introduced in api 17

ahmedre:
jazakAllah khairan

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
ozbek: It's `NotificationCompat` after all. It should have handled such cases internally and say _"Oh we got `setShowWhen(false)`, but we are running on an old platform. Let's replace it with `setWhen(0)`"_. But somehow it doesn't do that :)

ozbek:
wa iyyakum

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
ahmedre: we should file it b.android.com

PR #614, opened 2016-01-11
Remove null notification icon check and empty titles

ozbek:
As mentioned [elsewhere](https://github.com/quran/quran_android/pull/610#issuecomment-170500613), _that is an edge case_.

Null `mNotificationIcon` in Android 4.4.4:
null-notification-icon_android-4 4 4

Null `mNotificationIcon` in Android 6.0.1:
null-notification-icon_android-6 0 1

Null `mNotificationIcon` in Android 6.0.1 lockscreen:
null-notification-icon_android-6 0 1-lockscreen

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
+++ b/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
@@ -1095,8 +1095,6 @@ public class AudioService extends Service implements OnCompletionListener,
       }
     }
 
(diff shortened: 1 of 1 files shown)


ahmedre:
jazakAllah khairan - we'll know when we push how many people hit this case - if it's a lot, we can re-add the code to gracefully handle it insha'Allah, otherwise, this is better because it keeps the code cleaner.

PR #616, opened 2016-01-13
[i18n] Arabic: fix spelling / Uzbek: add reciters' names

ozbek:


app/src/main/res/values-ar/readers.xml
```diff
@@ -15,7 +15,7 @@
     <string name="qari_hani_rifai">هاني الرفاعي</string>
     <string name="qari_husary_mujawwad">الحصري مجود</string>
     <string name="qari_hudhayfi">الحذيفي</string>
-    <string name="qari_muaiqly">ماهر المعيلقي</string>
+    <string name="qari_muaiqly">ماهر_المعيقلي</string>
```
ahmedre: why did you add an `_` here?

ozbek: Sorry, underscore was unintentional. I only wanted to correct the sequence
of letters.

ahmedre: excellent catch, may Allah reward you greatly!

PR #624, opened 2016-01-25
Data fix for naskh

asimmohiuddin:
I made an error while making the last data fix. This one will fix that error.
The error causes incorrect audio to play on the last pages of juz 28


ahmedre:
jazakAllah khairan

PR #625, opened 2016-01-26
French update

abdullahibnnadjo:
As salâmu `alaykum wa rahmatuLlahi wa barakatuh,

Bâraka Allahu fikum for having made this application open source again !

Little update of the French language here

By the way, can I translate Play Store's description of the app ?

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,7 +3,7 @@ quran android contributors
 
 ### code
 
-* [Asim Mohiuddin](https://github.com/asimmohiuddin) - images, gapless audio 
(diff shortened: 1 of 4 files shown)


ahmedre:
jazakAllah khairan! sure, if you can send me a description, i can add it for French insha'Allah.

ozbek:
> if you can send me a description

IMHO, this is one of the main reasons why we should move to third-party* translation platforms for localizations. That'd enable translators easily access and translate the app strings as well as Play Store descriptions (without having to ask about it). 

* Self hosted platforms require maintenance and extra time.

PR #631, opened 2016-02-06
Update README.md

asimmohiuddin:
Ahmed's name was present when we used strings.xml for credits but when we moved to README.md and CONTRIBUTORS.md it was no longer present


ahmedre:
jazakAllah khairan :)

PR #649, opened 2016-04-24
Update deprecated methods

hakimrie:
Update deprecated with new ones or using ContextCompat


ahmedre:
jazakAllah khairan - test failures are because the android plugin is now out of date, so we need to update it. on an unrelated note, am surprised that our current styles file isn't putting the imports in alphabetical order, we should make sure to fix that insha'Allah.

hakimrie:
waiyyak, Alhamdulillah. I am sorry, I forgot to include gradle version update into this pull request.
thank you for accepting the pull request.

PR #655, opened 2016-05-07
Dutch Translation for Quran Android

abdelouali:


ozbek:
This seems to be failing because of `classpath 'com.android.tools.build:gradle:2.0.0-beta6'` in `build.gradle`...

abdelouali:
I will let @ahmedre  take care of this.  If he has time

ahmedre:
yep, the error is unrelated to the patch - basically, because we're using alpha/beta Android plugin versions, whenever a new version comes out, the previous one stops working (seemingly intentionally) - so we'll just have to update the plugin version to make tests pass again (and/or switch to a non-beta plugin version).

abdelouali:
Allright, thanks for the info.

PR #669, opened 2016-07-21
 Update strings.xml for Indonesian language

habibr:
Replaced several english words that are commonly used by indonesian users for corresponding indonesian language words that are equally commonly used or are getting popular, such as "Unduh" for Download and "Berkas" for File.

Rephrasing and redactional changes to current indonesian translation to reflect the more exact meaning of the english prompts and directions.

Re-order strings in indonesian translation to the same order as the english. Complete all string translations which were not translated in previous file.


ahmedre:
awesome, jazakAllah khairan - i am merging this, but if it's not too much trouble, could you remove all the strings that have `translateable=false` (to be similar to all the other translations as well which don't have these?)

PR #670, opened 2016-07-22
Update strings.xml for Indonesian language

habibr:
remove all occurrence of "translatable=false" as suggested by Ahmed


ahmedre:
jazakAllah khairan, but i meant remove those strings (not just the `translatable=false` piece

ahmedre:
this is already merged now, jazakAllah khairan again!

PR #675, opened 2016-08-10
fixed #667 retain scroll position when device orientation change

### LINKED ISSUE #667: remember scroll position on orientation change in translation
if you scroll in portrait mode to an ayah and switch to landscape, it returns to the top. we should at least return to the top of the ayah that was last being read.

hakimrie:
retain scroll position on translation page when device orientation change.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TranslationFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TranslationFragment.java
@@ -20,6 +20,7 @@ import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.ViewTreeObserver;
 
 import java.util.Set;
 
@@ -29,6 +30,8 @@ public class TranslationFragment extends Fragment
 
   private static final String SI_PAGE_NUMBER = "SI_PAGE_NUMBER";
   private static final String SI_HIGHLIGHTED_AYAH = "SI_HIGHLIGHTED_AYAH";
+  private static final String SI_PERCENT_VERTICAL_SCROLL_POSITION =
+      "SI_PERCENT_VERTICAL_SCROLL_POSITION";
 
   private int mPageNumber;
   private int mHighlightedAyah;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/fragment/TranslationFragment.java
```diff
@@ -39,6 +43,15 @@
...
   private QuranSettings mQuranSettings;
   private boolean mJustCreated;
+  private float mPercentLastScrollPosition;
+  private ViewTreeObserver.OnGlobalLayoutListener layoutListener =
+      new ViewTreeObserver.OnGlobalLayoutListener() {
+        @Override
+        public void onGlobalLayout() {
+          mTranslationView.scrollTo(0,
+              Math.round(mPercentLastScrollPosition * getTranslationViewHeight()));
+        }
```
ahmedre: i'd actually remove it here also

ahmedre:
my question to you is when this would matter - i guess only for translations, or also for when you accidentally move to portrait and back to landscape? if so, can we conditionally do this for translations (since in image mode in portrait, the entire screen fits the image, and only in landscape is it scrollable).

depending on this, if for translations, if we can figure out the first ayah on the screen easily (i have to look at the code, can't remember on top of my head), that might be better than a percentage since the percentage's meaning maybe different in portrait/landscape.

sorry for taking so long to reply to this

hakimrie:
@ahmedre yes, this implementation is just for translations ( as defined in issue #667 ).
because current `TranslationView` is a `ScrollView`, I can't think any other way to save current position beside by using scroll listener (it will be different if we use `ListView/RecyclerView`).

using percentage make it possible to adjust scroll position to the right position. since the width/height of the screen different between landscape and portrait, we can't use absolute scroll position in portrait and applied it in landscape mode.

ahmedre:
makes sense, jazakAllah khairan - i misread the patch the first time - looks good to me, just requesting that you remove the listener sooner as per the comment and then can merge.

ahmedre:
(as an aside, that's another good point - in the future, this should use a `RecyclerView` instead of manually adding stuff to a scrolling window)

hakimrie:
somehow, remove globallayoutlistener immediately after calling `scrollTo` didn't work, I think it because ayat/translation is added asynchronously using `TranslationTask`. I am thinking to refactor translationview to use `RecyclerView` instead of `ScrollView`. in the meantime I will close this PR.

ahmedre:
awesome, looking forward to it, jazakAllah khairan!

PR #687, opened 2016-09-23
Viewing new rub3 permenantly instead of just poping it up

hulk9914:
Asalamu 3alaikum,

**The current situation:**
When the user reaches a new rub3, a toast will pop up showing this rub3 number. If the user took some time reading the current page and wanted to know in which rub3 he was, he has to swipe left and then swipe right to show the toast again and I think that this is not user-friendly.

**The suggested change:**
There is a space in the top middle of the page, so why we don't show the current rub3 in this space upon reaching it.

Wa jazakum allaho khairan.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
@@ -320,8 +320,8 @@ public class TabletFragment extends Fragment
           if (mRightImageView != null && mLeftImageView != null) {
             mRightImageView.setPageBounds(rect[0]);
             mLeftImageView.setPageBounds(rect[1]);
-            if (mOverlayText) {
-              Context context = getContext();
+            Context context = getContext();
+            if (mOverlayText && isAdded()) {
               String suraText = QuranInfo.getSuraNameFromPage(context, mPageNumber - 1, true);
               String juzText = QuranInfo.getJuzString(context, mPageNumber - 1);
               String pageText = QuranUtils.getLocalizedNumber(context, mPageNumber - 1);
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
@@ -319,16 +319,18 @@ public class TabletFragment extends Fragment
           if (mRightImageView != null && mLeftImageView != null) {
             mRightImageView.setPageBounds(rect[0]);
             mLeftImageView.setPageBounds(rect[1]);
-            Context context = getContext();
-            if (mOverlayText && context != null) {
+            if (mOverlayText) {
+              Context context = getContext();
               String suraText = QuranInfo.getSuraNameFromPage(context, mPageNumber - 1, true);
               String juzText = QuranInfo.getJuzString(context, mPageNumber - 1);
               String pageText = QuranUtils.getLocalizedNumber(context, mPageNumber - 1);
-              mRightImageView.setOverlayText(suraText, juzText, pageText);
+              String rub3Text =HighlightingImageView.displayRub3(context,mPageNumber - 1);
+              mRightImageView.setOverlayText(suraText, juzText, pageText, rub3Text);
               suraText = QuranInfo.getSuraNameFromPage(context, mPageNumber, true);
               juzText = QuranInfo.getJuzString(context, mPageNumber);
               pageText = QuranUtils.getLocalizedNumber(context, mPageNumber);
-              mLeftImageView.setOverlayText(suraText, juzText, pageText);
+              rub3Text =HighlightingImageView.displayRub3(context,mPageNumber);
+              mLeftImageView.setOverlayText(suraText, juzText, pageText, rub3Text);
             }
           }
         }
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
@@ -27,6 +27,7 @@ import com.quran.labs.androidquran.ui.PagerActivity;
 import com.quran.labs.androidquran.ui.helpers.AyahSelectedListener;
 import com.quran.labs.androidquran.ui.helpers.AyahTracker;
 import com.quran.labs.androidquran.ui.helpers.HighlightType;
+import com.quran.labs.androidquran.ui.helpers.QuranDisplayHelper;
 import com.quran.labs.androidquran.ui.helpers.QuranPageWorker;
 import com.quran.labs.androidquran.ui.util.ImageAyahUtils;
 import com.quran.labs.androidquran.ui.util.PageController;
@@ -324,12 +325,12 @@ public class TabletFragment extends Fragment
               String suraText = QuranInfo.getSuraNameFromPage(context, mPageNumber - 1, true);
               String juzText = QuranInfo.getJuzString(context, mPageNumber - 1);
               String pageText = QuranUtils.getLocalizedNumber(context, mPageNumber - 1);
-              String rub3Text =HighlightingImageView.displayRub3(context,mPageNumber - 1);
+              String rub3Text =QuranDisplayHelper.displayRub3(context,mPageNumber - 1);
               mRightImageView.setOverlayText(suraText, juzText, pageText, rub3Text);
               suraText = QuranInfo.getSuraNameFromPage(context, mPageNumber, true);
               juzText = QuranInfo.getJuzString(context, mPageNumber);
               pageText = QuranUtils.getLocalizedNumber(context, mPageNumber);
-              rub3Text =HighlightingImageView.displayRub3(context,mPageNumber);
+              rub3Text = QuranDisplayHelper.displayRub3(context,mPageNumber);
               mLeftImageView.setOverlayText(suraText, juzText, pageText, rub3Text);
             }
           }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -166,9 +168,10 @@ public void adjustNightMode() {
     String suraText = null;
     String juzText = null;
     String pageText = null;
+    String rub3Text = null;
   }
 
-  public void setOverlayText(String suraText, String juzText, String pageText) {
+  public void setOverlayText(String suraText, String juzText, String pageText,String rub3Text) {
```
ozbek: space

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -178,6 +181,7 @@ public void setOverlayText(String suraText, String juzText, String pageText) {
     mOverlayParams.suraText = suraText;
     mOverlayParams.juzText = juzText;
     mOverlayParams.pageText = pageText;
+    mOverlayParams.rub3Text=rub3Text;
```
ozbek: space

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -245,6 +249,13 @@ private void overlayText(Canvas canvas, Matrix matrix) {
     canvas.drawText(mOverlayParams.pageText,
         getWidth() / 2.0f, mOverlayParams.bottomBaseline,
         mOverlayParams.paint);
+     // Write the current rub3 text at the top middle of the page
+     if (!mOverlayParams.rub3Text.equals("")) {
+          mOverlayParams.paint.setTextAlign(Align.CENTER);
+          canvas.drawText(mOverlayParams.rub3Text,
+                  getWidth() / 2.0f, mOverlayParams.topBaseline,
+                  mOverlayParams.paint);
+      }
```
ozbek: space: does not seem to align with other lines

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranDisplayHelper.java
```diff
@@ -39,6 +39,29 @@ public static Response getQuranPage(OkHttpClient okHttpClient,
     return response;
   }
 
+  // same logic used in displayMarkerPopup method
+  public static String displayRub3(Context context, int page)
+  {
```
ozbek: Please use the same line as method declaration for opening bracket

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranDisplayHelper.java
```diff
@@ -39,6 +39,29 @@ public static Response getQuranPage(OkHttpClient okHttpClient,
     return response;
   }
 
+  // same logic used in displayMarkerPopup method
```
ozbek: I think, for this comment to make sense,  `displayRub3()` needs to come below/after `displayMarkerPopup()`

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
```diff
@@ -319,16 +320,18 @@ protected void onPostExecute(RectF[] rect) {
           if (mRightImageView != null && mLeftImageView != null) {
             mRightImageView.setPageBounds(rect[0]);
             mLeftImageView.setPageBounds(rect[1]);
-            Context context = getContext();
-            if (mOverlayText && context != null) {
```
ozbek: +1! 
There is no point in checking if context is null. If it is null, then something is seriously broken beyond repair.

ahmedre: i intentionally added that null check because we were having crashes due to this, so please re-add this back.

because this is in `onPostExecute`, we can easily have a case where the fragment is killed before the `AsyncTask` finishes - when the `AsyncTask` is done, there is no `context` because the `Fragment` is gone, so we should just bail - by not doing this check, we'll crash in the next few lines where we try to use the context to get sura names, etc.

ahmedre:
i am not against this patch conceptually, but have a few suggestions (aside from addressing the comments) - i think this will look strange in the top middle. my suggestion instead is either:
a. merge it with the juz' text at the top right or
b. move the page number to the left (or right, depending on the side of the page the page is on), and on the opposite side, put the hizb info.

ahmedre:
(oh, and most importantly, jazakumAllah khairan for doing this!)

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
ozbek: I see. 

I think, for that scenario, checking if fragment is being removed ([`isRemoving()`](https://developer.android.com/reference/android/app/Fragment.html#isRemoving())) or if it is visible ([`isVisible()`](https://developer.android.com/reference/android/app/Fragment.html#isVisible())) would fit better.

ahmedre: good call, i agree. in `QuranPageFragment`, we're checking `isAdded()` and that seems to be working fine (no crash there), so am good with wrapping the entire block with an `isAdded()` check.

hulk9914:
First, I'm sorry about not following some styling rules.
Second, @ahmedre I think option "a" is better and I'm working on it.
JazkomoAllah Khairan.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
```diff
@@ -319,16 +320,18 @@ protected void onPostExecute(RectF[] rect) {
...
+            if (mOverlayText) {
```
ozbek: Please add `isAdded()` check.

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranDisplayHelper.java
```diff
@@ -73,6 +73,29 @@ public static long displayMarkerPopup(Context context, int page,
     return System.currentTimeMillis();
   }
 
+  // same logic used in displayMarkerPopup method
+  public static String displayRub3(Context context, int page){
+    int rub3 = QuranInfo.getRub3FromPage(page);
+    int hizb = (rub3 / 4) + 1;
+    StringBuilder sb = new StringBuilder();
+    sb.append(" , ");
+    if (rub3 == -1) {
```
ozbek: Can we do this check earlier (right after line 78)?

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranDisplayHelper.java
```diff
@@ -73,6 +73,29 @@ public static long displayMarkerPopup(Context context, int page,
...
+      return "";
+    }
+    int remainder = rub3 % 4;
```
ozbek: For the sake of readability (just a suggestion):

    switch(rub3%4) {
        case 1:
            sb.append(context.getString(R.string.quran_rob3)).append(' ');
            break;
        case 2:
            sb.append(context.getString(R.string.quran_nos)).append(' ');
            break;
        case 3:
            sb.append(context.getString(R.string.quran_talt_arb3)).append(' ');
            break;
        default: // case 0
            // do nothing
    }

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
hulk9914: oh, sorry I totally forgot that.
now added.

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranDisplayHelper.java
hulk9914: Done.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
```diff
@@ -320,15 +321,17 @@ protected void onPostExecute(RectF[] rect) {
             mRightImageView.setPageBounds(rect[0]);
             mLeftImageView.setPageBounds(rect[1]);
             Context context = getContext();
-            if (mOverlayText && context != null) {
+            if (mOverlayText && isAdded()) {
```
ozbek: +1! Now, since context is not used outside, you could move `Context context = getContext();` to inside the brackets (your original change). Sorry for sounding like nitpicking :(

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
ozbek: spacing is not fixed yet

hulk9914: Now fixed

app/src/main/java/com/quran/labs/androidquran/ui/fragment/TabletFragment.java
hulk9914: May Allah rewards you for your great effort in improving this app.

ahmedre:
merging to unblock, jazakAllah khairan.
if you get a chance, please also do @hakimrie's suggestion (in a new patch).

PR #706, opened 2016-11-21
.editorconfig

hafah:
adding an .editorconfig file to the repo so that new contributors will automatically have their formatting setup correctly.

It's built into Android Studio. Settings->Editor->Code Style then Check "Enable EditorConfig".

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,22 +5,5 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
-
-# Matches the exact files either package.json or .travis.yml
-[{package.json,.travis.yml}]
-indent_style = space
-indent_size = 2
-
-# Tab indentation (no size specified)
-[Makefile]
-indent_style = tab
-
-# Indentation override for all java files
-[*.java]
-indent_style = space
-indent_size = 2
-
-# 2 space indentation
-[*.sh]
 indent_style = space
 indent_size = 2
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+insert_final_newline = true
+
+# Matches the exact files either package.json or .travis.yml
+[{package.json,.travis.yml}]
+indent_style = space
+indent_size = 2
+
+# Tab indentation (no size specified)
+[Makefile]
+indent_style = tab
```
ozbek: Why tab here?

hafah: Great catch! I don't know why I left a code style specific to MakeFile. I'll remove it.

ahmedre:
jazakAllah khairan, never knew about this - looks like this is respected by other IDEs and editors too, which is a nice touch.

can this work okay in addition to quran_android-code_style.xml? the benefit of that one is, for IntelliJ/Android Studio at least, is that it gives us other things (what order should imports be in, how many spaces before and after an array declaration, etc).

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
```
ahmedre: can we also add:
`charset = utf-8`?

ahmedre:
jazakAllah khairan

PR #707, opened 2016-11-21
.editorconfig in the readme

hafah:
sorry for not putting this in the ".editorconfig" pull request #706

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ contributors:
 * [Rehab Mohamed](https://twitter.com/hams_rrr)
 * [Ahmed Essam](https://twitter.com/neo_4583)
 
-
 Code Style
 ------------------------
 General [Android code style guidelines](https://source.android.com/source/code-style.html) apply, with the exception of following indent sizes:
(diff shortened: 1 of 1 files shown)


ahmedre:
jazakAllah khairan - per discussion point on #706, can we leave the existing instructions and add these as additional instructions?

hafah:
We can, but more than implementing .editorconfig I want to have the README file only for information not for interactions and how-to's. I feel like the README file is intented for those who want to "read up" on everything not for those who want to "set up" everything.

So this depends on what other people think.

P.S. When I have time I'll check if this can replace the .xml file like you asked [in this comment](https://github.com/quran/quran_android/pull/706#issuecomment-262182203)

ahmedre:
> I feel like the README file is intented for those who want to "read up" on everything not for those who want to "set up" everything.

agreed - i recently made some changes, can you re-add this (but keep reference to using the xml file as our code styles in?)

from what i read on the editor config website, i don't think it can fully replace the xml document, but it will be very helpful for people using non-IntelliJ/Android Studio IDEs.

ahmedre:
this unfortunately has a bad side-effect of breaking our current indentation rules. for example, tabs are correctly set to 2 spaces, but nested tabs (i.e. if a line of code splits on several lines), we're using 4 spaces instead of 2 in IntelliJ/Android Studio, but this forces only 2 spaces.

ahmedre:
update: apparently intellij does support it (even though editorconfig itself doesn't), so i'll just add this to .editorconfig - https://github.com/JetBrains/intellij-community/commit/570f7cba223e02032e810731d44d8b26a7258dfb

Issue #704, opened 2016-11-15
Reorganize Settings Menu


ahmedre:

ahmedre:
The settings menu is getting a bit unwieldy.

Thoughts:
1. maybe move "Advanced Options" into a separate screen
2. maybe move "Display Options" into a separate screen
3. move "Volume Key Navigation" under Reading preferences

"maybe" because I could be convinced either way - open to feedback.

ahmedre:
anyone should feel free to work on this and send a PR

ozbek:
> move "Volume Key Navigation" under Reading preferences

+1... OK, no "maybe" there, but still I voted :)

ahmedre:
lol, everything should be considered a maybe :)

ozbek:
> maybe move "Display Options" into a separate screen

Regarding "Display Options", this seems to overlap with "Reading Preferences". For example, "New background" or "Lock screen orientation" could have been in "Reading Preferences", or "Show page info" or "Highlight bookmarks" could be put to "Display Options". 

Only "Arabic mode" does not seem to fit to "Reading Preferences" as that's purely about the UI. Is it not time to remove that option altogether? Android supports Arabic natively now.

ahmedre:
i am good with your PR - but just wanted to comment on the Arabic setting - i personally use this (despite my phone being in English as the default locale) - i think it probably gets a lot of use from:

1. people who speak Arabic fluently but usually converse in English - so they leave their phones in English and just enable Arabic for Islamic apps when the option is there
2. people who are learning Arabic

so i personally would be sad if this went away :(

PR #711, opened 2016-11-27
Split the setting advanced options into a new preferences screen This commit solves Issue Reorganize Settings Menu #704

mhussien86:


ahmedre:
jazakAllah khairan - unfortunately, this has the problem of losing the toolbar at the top:
screenshot-2016-11-28_00 52 05 108

that's because of how we are working around limitations and supporting a toolbar on previous api versions. one way to do this is to add a new activity that just loads those advanced options, and connect the click from the main activity to starting the new activity.

mhussien86:
Okay i will be working on that .. thanks :)

PR #713, opened 2016-11-28
Use recent pages size to correctly get quantity string

ozbek:
This also removes two unnecessary local variables, `boolean showLastPage` and `int length`.


ahmedre:
jazakAllah khairan

PR #714, opened 2016-11-28
Split setting screen into options and advanced options issue  #704

mhussien86:
This commit is solving the reorganizing of the setting screen into options and advanced options, change in the category name of Advanced settings maybe needed in the future.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -52,7 +52,7 @@
         android:theme="@style/QuranToolBar"
         android:configChanges="keyboardHidden|orientation|screenSize"/>
     <activity
-        android:name=".AdvancedQuranPreferenceActivity"
+        android:name=".QuranAdvancedPreferenceActivity"
         android:theme="@style/QuranToolBar"
         android:configChanges="keyboardHidden|orientation|screenSize"/>
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -52,6 +52,11 @@
         android:theme="@style/QuranToolBar"
         android:configChanges="keyboardHidden|orientation|screenSize"/>
     <activity
+        android:name=".AdvancedQuranPreferenceActivity"
+        android:theme="@style/QuranToolBar"
+        android:configChanges="keyboardHidden|orientation|screenSize"/>
+
+    <activity
         android:name=".SearchActivity"
         android:launchMode="singleTop">
       <intent-filter>
--- a/app/src/main/java/com/quran/labs/androidquran/AdvancedQuranPreferenceActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/AdvancedQuranPreferenceActivity.java
@@ -1,0 +1,104 @@
+package com.quran.labs.androidquran;
+
+import android.Manifest;
+import android.app.Fragment;
+import android.app.FragmentManager;
+import android.content.pm.PackageManager;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.v4.app.ActivityCompat;
+import android.support.v7.app.ActionBar;
+import android.support.v7.widget.Toolbar;
+import android.view.MenuItem;
+import android.widget.Toast;
+
+import com.quran.labs.androidquran.service.util.PermissionUtil;
+import com.quran.labs.androidquran.ui.QuranActionBarActivity;
+import com.quran.labs.androidquran.ui.fragment.QuranAdvancedSettingsFragment;
+import com.quran.labs.androidquran.util.AudioManagerUtils;
+import com.quran.labs.androidquran.util.QuranSettings;
+
+public class AdvancedQuranPreferenceActivity extends QuranActionBarActivity {
+
+  private static final String SI_LOCATION_TO_WRITE = "SI_LOCATION_TO_WRITE";
+  private static final int REQUEST_WRITE_TO_SDCARD_PERMISSION = 1;
+
+  private String mLocationToWrite;
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    ((QuranApplication) getApplication()).refreshLocale(this, false);
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.preferences);
+
+    final Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
+    toolbar.setTitle(R.string.menu_settings);
+    setSupportActionBar(toolbar);
+    final ActionBar ab = getSupportActionBar();
+    if (ab != null) {
+      ab.setDisplayHomeAsUpEnabled(true);
+    }
+
+    AudioManagerUtils.clearCache();
+
+    if (savedInstanceState != null) {
+      mLocationToWrite = savedInstanceState.getString(SI_LOCATION_TO_WRITE);
+    }
+
+    final FragmentManager fm = getFragmentManager();
+    final Fragment fragment = fm.findFragmentById(R.id.content);
+    if (fragment == null) {
+      fm.beginTransaction()
+          .replace(R.id.content, new QuranAdvancedSettingsFragment())
+          .commit();
+    }
+  }
+
+  @Override
+  protected void onSaveInstanceState(Bundle outState) {
+    if (mLocationToWrite != null) {
+      outState.putString(SI_LOCATION_TO_WRITE, mLocationToWrite);
+    }
+    super.onSaveInstanceState(outState);
+  }
+
+  @Override
+  public boolean onOptionsItemSelected(MenuItem item) {
+    if (item.getItemId() == android.R.id.home) {
+      finish();
+      return true;
+    }
+    return super.onOptionsItemSelected(item);
+  }
+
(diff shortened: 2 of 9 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -52,6 +52,11 @@
         android:theme="@style/QuranToolBar"
         android:configChanges="keyboardHidden|orientation|screenSize"/>
     <activity
+        android:name=".AdvancedQuranPreferenceActivity"
```
ozbek: IMHO, this should follow existing file naming convention. So, probably, should be named as `QuranAdvancedPreferenceActivity`.

app/src/main/java/com/quran/labs/androidquran/AdvancedQuranPreferenceActivity.java
```diff
@@ -0,0 +1,104 @@
...
+  private String mLocationToWrite;
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    ((QuranApplication) getApplication()).refreshLocale(this, false);
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.preferences);
+
+    final Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
+    toolbar.setTitle(R.string.menu_settings);
```
ozbek: Please use "Advanced settings" title.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/QuranAdvancedSettingsFragment.java
```diff
@@ -0,0 +1,462 @@
...
+import rx.Observable;
+import rx.Subscriber;
+import rx.Subscription;
+import rx.android.schedulers.AndroidSchedulers;
+import rx.functions.Action1;
+import rx.functions.Func1;
+import rx.schedulers.Schedulers;
+import timber.log.Timber;
+
+public class QuranAdvancedSettingsFragment extends PreferenceFragment implements
```
ozbek: If we are to use an additional fragment, is it possible to just have one original activity and let it utilize two fragments?

app/src/main/res/xml/quran_preferences.xml
```diff
@@ -147,29 +147,16 @@
   </PreferenceCategory>
 
   <PreferenceCategory
```
ozbek: Is it possible to handle `PreferenceCategory` clicks and launch advanced settings thereof, and get rid of additional `Preference` below?

ozbek:
Also, while you are at it, please consider implementing all the changes mentioned in #704.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/QuranAdvancedSettingsFragment.java
```diff
@@ -0,0 +1,462 @@
...
+  @Override
+  public void onPause() {
+    mIsPaused = true;
+    getPreferenceScreen().getSharedPreferences()
+        .unregisterOnSharedPreferenceChangeListener(this);
+    super.onPause();
+  }
+
+  @Override
+  public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String s) {
```
ahmedre: remove this

app/src/main/java/com/quran/labs/androidquran/ui/fragment/QuranAdvancedSettingsFragment.java
```diff
@@ -0,0 +1,462 @@
...
+import rx.Subscriber;
+import rx.Subscription;
+import rx.android.schedulers.AndroidSchedulers;
+import rx.functions.Action1;
+import rx.functions.Func1;
+import rx.schedulers.Schedulers;
+import timber.log.Timber;
+
+public class QuranAdvancedSettingsFragment extends PreferenceFragment implements
+    SharedPreferences.OnSharedPreferenceChangeListener {
```
ahmedre: you don't need the `implements SharedPreferences.OnSharedPreferenceChangeListener` here.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/QuranAdvancedSettingsFragment.java
ahmedre: let's keep it as 2 and clean it up later since this is mostly a straight forward move of code - perhaps in another PR if it makes sense?

app/src/main/res/xml/quran_preferences.xml
ahmedre: in an ideal world, the patch would have been as simple as Mohamed's original patch where he'd just add `PreferenceScreen` around it - unfortunately though it doesn't work. maybe it would work with support-v7 fragments, but we can try that later?

app/src/main/AndroidManifest.xml
ahmedre: +1

ahmedre:
re #704 agree, but split them into separate patches - let's get this one merged first and then if you're cool with continuing, would love more PRs about this insha'Allah?

app/src/main/java/com/quran/labs/androidquran/AdvancedQuranPreferenceActivity.java
mhussien86: Thiught about using the exsting string as we will need a localization String for any new String will do this.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/QuranAdvancedSettingsFragment.java
mhussien86: Agree for now, just to avoid more checks and condition.

app/src/main/java/com/quran/labs/androidquran/AdvancedQuranPreferenceActivity.java
ahmedre: use `R.string.prefs_category_advanced` for now

app/src/main/res/xml/quran_preferences.xml
mhussien86: Can you elaborate more ? do we need to change something here ?

ahmedre: i think @ozbek was saying, since `PreferenceCategory extends Preference`, instead of having the `PreferenceCategory` have a nested `Preference` for advanced settings, can you just make `PreferenceCategory` handle advanced settings without adding a child in xml?

i.e. instead of what you have now:

```xml
<PreferenceCategory android:key="">
   <Preference android:key="">
        <intent android:targetPackage="" />
   </Preference>
</PreferenceCategory>
```

instead, would it work if you move everything on the `Preference` up to the `PreferenceCategory` instead and remove the preference? i.e.

```xml
<PreferenceCategory android:key="">
    <intent android:targetPackage="" />
</PreferenceCategory>
```

ahmedre: but i don't think this will work because the preference category provides a strange ui on its own - so i think you can just leave this for now.

ahmedre:
jazakAllah khairan - +1 from me, but will wait for @ozbek's +1 before merging.

app/src/main/res/xml/quran_advanced_preferences.xml
```diff
@@ -0,0 +1,31 @@
...
+        android:title="@string/prefs_app_location_title"/>
+
+    <com.quran.labs.androidquran.ui.preference.QuranPreference
+        android:key="@string/prefs_send_logs"
+        android:title="@string/prefs_send_logs_title"
+        android:summary="@string/prefs_send_logs_summary"/>
+  </PreferenceCategory>
+</PreferenceScreen>
+
+
```
ozbek: Unnecessary empty line

app/src/main/res/xml/quran_preferences.xml
```diff
@@ -147,29 +147,16 @@
...
-        android:summary="@string/prefs_send_logs_summary"/>
-  </PreferenceCategory>
-
+<Preference
+      android:key="@string/prefs_path"
+      android:title="@string/prefs_category_advanced">
+    <intent
+        android:targetPackage="com.quran.labs.androidquran"
+        android:targetClass="com.quran.labs.androidquran.QuranAdvancedPreferenceActivity"
+        android:action="com.quran.labs.androidquran.PREFS_ONE" />
```
ozbek: Why do we need this action?

ozbek:
> also about the one activity with two fragments won't it be a one level of navigation what do you think i prefer advanced options to be in a second level of navigation.

That's basically what I am asking about :)

Is it possible to have your desired behaviour without adding `QuranAdvancedPreferenceActivity`?

mhussien86:
Yes possible but not cleaner it will be with some hacks, like sending string value from setting fragment and restart setting activity with string value to replace setting fragment with advanced settings fragment ! , the current we have is more separated and better.

ozbek:
OK, please cleanup the unused code. Thank you :)

ahmedre:
i'll go ahead and merge and then please send another PR removing whatever should be removed. jazakAllah khairan for doing this @mhussien86 !

PR #715, opened 2016-11-29
Use same text as in the menu to jump last page

ozbek:
This is to have a consistent UI text accross the app. When there is only one recently read page, the title will be the same as the jump menu title in the action bar.

This is a conceptual change. Please feel free to argue against :)


ahmedre:
jazakAllah khairan

PR #716, opened 2016-11-29
Add translation for "Recent pages"

ozbek:
Normally, `plurals` require special approach for each locale. But here, we do not care about actual number of recent pages except for one case: 1 recent page (last page) or more. So, instead of translating `plurals`, it is more convenient to just use the `string`.

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -281,13 +281,7 @@
     <item quantity="many">آخر المتصفحات</item>
   </plurals>
 
-  <plurals name="bookmark_tag_deleted">
-    <item quantity="zero">لم يتم حذف عناصر</item>
-    <item quantity="one">تم حذف عنصر واحد</item>
-    <item quantity="two">تم حذف عنصران</item>
-    <item quantity="few">تم حذف %d عناصر</item>
-    <item quantity="many">تم حذف %d عنصر</item>
-  </plurals>
+  <string name="recent_pages">تم حذف عنصران</string>
   <string name="undo">تراجع</string>
 
 </resources>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -339,9 +339,10 @@
   </plurals>
   <string name="undo">Undo</string>
 
-  <plurals name="recent_pages">
+  <string name="recent_pages">Recent pages</string>
+  <plurals name="recent_pages" translatable="false">
```
ahmedre: can you either remove `translatable="false"` or remove the plurals from `values-ar` (should this be okay in locales that require plurals of zero, one, two, ...?)

also, super minor nitpick - can we name the plural different from the string?

app/src/main/res/values-ar/strings.xml
```diff
@@ -281,13 +281,7 @@
     <item quantity="many">آخر المتصفحات</item>
   </plurals>
 
-  <plurals name="bookmark_tag_deleted">
```
ahmedre: wrong one removed

app/src/main/res/values/strings.xml
ozbek: > should this be okay in locales that require plurals of zero, one, two, ...?

That's where I got the idea. For instance, in Arabic and Russian, this requires `%d` argument for plurals. As we do not care about actual numbers (in fact, we are not feeding any numbers for format args), that obviously does not work. Also, we do not show the last page / recent pages header when the amount is zero. So, no, this won't be an issue (hopefully:):

 - Zero --> not displated
 - Only one --> Last page
 - More than one --> Recent pages

The other are done in the second commit.

ahmedre:
jazakAllah khairan

PR #724, opened 2016-12-06
RxJava2

ahmedre:


ahmedre:
this patch is a work in progress. the biggest thing remaining is making sure we're properly handling nulls now that RxJava2 no longer accepts them.

ozbek:
Whoa, lots' of change! :+1:

PR #726, opened 2016-12-06
Show translation info even if translator has no localized name

ozbek:
I don't know why translation info was not shown if translator did not have a localized name :)


ahmedre:
jazakAllah khairan

PR #727, opened 2016-12-07
Fix bugs with recent pages

### LINKED ISSUE #725: bug with RecentPages
while developing, i noticed that sometimes, `onPageChanged` is called before `bind` is called in the `RecentPagePresenter` - consequently, this offsets the page by one and results in things being wrong.

ahmedre:
This patch fixes #725. It exposes an Observable that observes every time
onPageSelected is called on a ViewPager. RecentPagePresenter subscribes
to this and uses this information to know how to update its recent page
state.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
@@ -105,9 +105,13 @@ import java.util.concurrent.Callable;
 
 import javax.inject.Inject;
 
+import io.reactivex.Observable;
+import io.reactivex.ObservableEmitter;
+import io.reactivex.ObservableOnSubscribe;
 import io.reactivex.Single;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.CompositeDisposable;
+import io.reactivex.functions.Cancellable;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -493,8 +496,29 @@ public void onPageSelected(int position) {
     downloadReceiver.setListener(this);
   }
 
-  private void saveRecentPage(int page) {
-    recentPagePresenter.onPageChanged(page);
+  public Observable<Integer> getViewPagerObservable() {
```
ahmedre: some "thinking out loud" thoughts:

i am thinking something like this should be used to also clean up other logic (our massive `onCreate` method in `PagerActivity` for example. thinking about splitting some of that logic into a `PagePresenter` of sorts, and that can also `getViewPagerObservable` and subscribe to it so as to handle changes there separately from here.

at that point, we can then decide to cache this Observable in a variable initialized in onCreate and calling `share` on it so we can have multiple observers to it (with only one `OnPageChangeListener`).

another thought:
should we split out presenters and some sort of "view presenters" - i.e. what's the right place for this method to live? it felt strange for this to be `getViewPager` instead and the work be done in the presenter.

PR #730, opened 2016-12-11
Save translator's localized name in db

### LINKED ISSUE #692: Uzbek translator's name is out of sync
Translations drop down:
screenshot_20161010-112759

Translations settings:
screenshot_20161010-112821

ozbek:
Fixes #692.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBHelper.java
+++ b/app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBHelper.java
@@ -4,21 +4,10 @@ import android.content.Context;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
 
-import timber.log.Timber;
-
 class TranslationsDBHelper extends SQLiteOpenHelper {
 
   private static final String DB_NAME = "translations.db";
   private static final int DB_VERSION = 2;
-  private static final String CREATE_TRANSLATIONS_TABLE =
-      "CREATE TABLE " + TranslationsTable.TABLE_NAME + "(" +
-          TranslationsTable.ID + " integer primary key, " +
-          TranslationsTable.NAME + " varchar not null, " +
-          TranslationsTable.TRANSLATOR + " varchar, " +
-          TranslationsTable.TRANSLATOR_FOREIGN + " varchar, " +
-          TranslationsTable.FILENAME + " varchar not null, " +
-          TranslationsTable.URL + " varchar, " +
-          TranslationsTable.VERSION + " integer not null default 0);";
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBHelper.java
```diff
@@ -4,40 +4,53 @@
...
+  }
+
+  @Override
+  public void onCreate(SQLiteDatabase db) {
+    db.execSQL(CREATE_TRANSLATIONS_TABLE);
+  }
+
+  @Override
+  public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+    if (newVersion <= oldVersion) {
```
ahmedre: strictly speaking, probably not needed since this would indicate a framework bug

app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBHelper.java
```diff
@@ -4,40 +4,53 @@
...
+      Timber.w("Can't downgrade from version %d to version %d", oldVersion, newVersion);
+      return;
+    }
+
+    if (oldVersion < 2) {
+      // a new column is added and columns are re-arranged
+      db.execSQL("DROP TABLE IF EXISTS " + TranslationsTable.TABLE_NAME);
```
ahmedre: the downside of this is that it will break translations until the person re-visits the translation screen to re-build the database.

so can we instead of this, either:
alter table add column, and run a query to set the default to translator for now

or

rename the existing table, create the new table, and insert into the new table from the old table, replacing that column with translator.

what do you think?

ozbek: Yeap, not enough reason to "break" translations. I will look into it...

ahmedre:
jazakAllah khairan, awesome - nitpicks (but i am merging anyway):
1. perhaps wrap the upgrade logic (create table, insert all, drop, rename) in a transaction (and fallback to just drop / create?)
2. i prefer a static field but i am ok with the method for getting the create statement (i would argue "performance" but in reality, it doesn't matter since 99.9% of the time, this won't be called - only on the first time we're upgrading from 1 to 0 (or when we make a new database)).

ozbek:
Regarding the second point, I had to create a method to not to duplicate create table statement. I will look into the point 1... jazakAllah khair.

ahmedre:
you can do it without a method if you change the order - rename the old one first, then create the new one with the correct name, then run your insert.

ozbek:
Sorry, I could not get around the point 1. I don't know what exactly you mean by _"wrap the upgrade logic in a transaction and fallback to just drop/create"_.

PR #736, opened 2016-12-13
Remember the last reading mode

### LINKED ISSUE #530: remember selection to see translations
when a user goes into translations, we should probably default them to translations by default.

ozbek:
This will enable going back to translation mode when jump to last page is selected from the action bar.

Fix #530.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/data/Constants.java
+++ b/app/src/main/java/com/quran/labs/androidquran/data/Constants.java
@@ -72,4 +72,5 @@ public class Constants {
   public static final String PREF_LOGS = "sendLogsKey";
   public static final String PREF_DID_PRESENT_PERMISSIONS_DIALOG =
       "didPresentStoragePermissionDialog";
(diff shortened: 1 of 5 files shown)


ozbek:
Please don't merge yet. This has side effect on recent pages.

ahmedre:
what is the side effect on recent pages? it might be my bug if so, because otherwise, this seems right (code wise anyway).

ozbek:
> what is the side effect on recent pages?

Once the mode is changed to translation, all entries go to translation mode (from sura list, bookmarks, etc) except from "Jump to" menu (this always opens in Quran mode). I think, I need more details on the "concept" of #530. Should we open the page in previous mode only from "Last page" menu? Or always?

ahmedre:
i am ok either way (i can see the argument for both).

ozbek:
if so, please go ahead and merge :)

ahmedre:
jazakAllah khairan

PR #737, opened 2016-12-14
Enable installing debug build along with release version

ozbek:
Sometimes, it is more handy to have an everyday app separately from the one used for development.


ozbek:
Thoughts on logo:
 - is it better to have "debug" text at the bottom?
 - maybe shorten the text to "dev"?

ahmedre:
awesome idea, and definitely agree.

regarding the logo - i'd vote either for D or Dev or just a different color, with app_name string being Quran Dev maybe?

ahmedre:
btw, one thing to keep in mind that debug applies to everything...
i.e:

madaniDebug
qaloonDebug
...

i don't think we need debug for each one but just pointing it out so it doesn't bite us without us noticing. thoughts?

ozbek:
> regarding the logo - i'd vote either for D or Dev or just a different color

+1 for Dev. Which color? Green? :)

> with app_name string being Quran Dev maybe?

`app_name` name is localized. I'd hate to create app name for every locale in debug folder, I think that defies the benefit of this change. Logo because that was the minimum change required that I could think of. 

> one thing to keep in mind that debug applies to everything

Yes, should work with every flavor except logo will be from main app (with Dev title).

app/src/debug/AndroidManifest.xml
```diff
@@ -5,8 +5,10 @@
           android:installLocation="auto">
 
   <application
+      android:icon="@drawable/debug_icon"
+      android:label="Quran Debug"
```
ozbek: This will only be visible in application manager in Settings app. Does not effect the title in launcher.

ahmedre:
green sounds good to me (ideally a slightly different color from the color we're using in Qaloon or Naskh, can't remember which one) - yeah don't bother with the strings change then.

app/build.gradle
```diff
@@ -72,6 +72,9 @@ android {
 
       debug {
          ext.enableCrashlytics = false
+         applicationIdSuffix ".debug"
+         resValue "string", "authority", applicationIdSuffix + '.data.QuranDataProvider'
+         resValue "string", "file_authority", applicationIdSuffix + '.fileprovider'
```
ahmedre: not sure if these are right. these should be whatever the package name is plus the suffix so that if you build qaloonDebug, you get the right package name for example (probably have to try it and see)

ozbek:
device-2016-12-15-150450

app/build.gradle
```diff
@@ -84,6 +78,14 @@ android {
       }
     }
 
+  applicationVariants.all { variant ->
```
ozbek: This follows indentation rules defined in editor config. Please let me know if it needs to be aligned with the existing style.

ahmedre: yours is correct, we should re-indent the whole file

ahmedre:
jazakAllah khairan

PR #739, opened 2016-12-17
Fix f/c in non-Madani build types

ozbek:
That's caused by 2987bbf0f806eca440081f63a3240055f8290945 because advanced settings intent was hard coded in xml.
This also adds a summary to advanced settings preference.


ahmedre:
jazakAllah khairan

PR #742, opened 2016-12-19
Fix search menu in bookmarks tab

ozbek:
Apparently, I had broken it at e719dc42f37c0f98e35ff0f1fc4601a76fc9b926 :( 
This is a semi-revert with indentation fix in home_menu.xml.

There is a workaround to put a divider between menu items as explained [here](http://stackoverflow.com/questions/33277124/how-to-add-dividers-between-specific-menu-items), but I am not sure if that's worth the change.


ozbek:
To be clear, `invalidateOptionsMenu()` is removing reference to search view. Hence, no action is getting triggered when search is selected.

ahmedre:
i wouldn't bother with the divider.
jazakAllah khairan

Issue #557, opened 2015-07-13
stream setting may not check if the file is downloaded


ahmedre:
i got an email bug report which seems to imply that when the "prefer streaming" setting is checked, that we aren't checking to see if we have the file locally first. we should do this if we aren't already.

mhussien86:
Need to be tested ? anyone worked on this ? are we still have this issue ?

ahmedre:
i haven't verified it myself yet, but i wouldn't be surprised if it were true.

ozbek:
@ahmedre, do you want to keep this open until the proper "rewrite" of AudioService?

ahmedre:
no that's ok - we'll close this given your fix

PR #743, opened 2016-12-20
Check if local copy is available prior to streaming from remote server

ozbek:
This is a quick/dirty fix for #557.


ahmedre:
this is a great fix until we rewrite it - i remember looking into it some time ago and saying, "this mostly needs to be rewritten" :(

jazakAllah khairan!

PR #744, opened 2016-12-21
Remove unnecessary arg in AudioUtils.getQariUrl()

ozbek:
Apparently, getQariUrl() is called with `true` as the second argument everywhere.


ahmedre:
jazakAllah khairan, great catch

PR #747, opened 2016-12-29
Only read intent if not saved instance state

ahmedre:
Because the code for reading values for the ViewPager always happened
after checking saved instance state, the values there would override the
saved instance state values. This fixes a bug where, upon opening a
page, switching to the other mode, and switching a few pages, you'd go
back to the mode you started in before switching (or, on tablet, it'd
happen right away when switching from portrait to landscape).


ahmedre:
cc @ozbek

PR #748, opened 2016-12-30
709 Refactored TranslationManager/Adapter

hosainnet:
- Unsubscribe from UnicastSubjects in TranslationManager (was causing crash the 2nd time you visit the translations view)
- Moved translation adapter listeners to ViewHolder's constructor for performance


hosainnet:
Will re-create on a clean branch

PR #749, opened 2016-12-30
709 refactor translation manager

hosainnet:
- Unsubscribe from UnicastSubjects in TranslationManager (was causing crash the 2nd time you visit the translations view)

-Moved translation adapter listeners to ViewHolder's constructor for performance


ahmedre:
jazakAllah khairan
one last thing actually - if you get a chance, may you add a press state to the rows? you can probably re-use some of the drawable xmls we're using for sura list / juz' list, etc

Issue #722, opened 2016-12-04
Please allow vertical scrolling while translation toolbar is visible


amynbe:
السلام عليكم
The KSU Ayat app has the correct behaviour here: 
- Translation covers text but you can still scroll up to show the verses at the bottom of the pages. 
- Moreover, the translation area itself shows all the verses of the page and is scrollable. 
- When playing recitation, the translation moves with the currently recited verse.
- Translation area has configurable height.

Can we please have the same bevahiour implemented in Quran Android? What's your view on the above points?

Wa jazakumullah kheiran

ahmedre:
wa3laikum alsalam,
jazakAllah khairan - let me address them one at a time:

> Translation covers text but you can still scroll up to show the verses at the bottom of the pages.

agree - this is planned insha'Allah

> Moreover, the translation area itself shows all the verses of the page and is scrollable.

we already have this - there are 2 translation "modes" - the full screen mode (tap the screen and tap the globe) and the panel mode (the one i am guessing you are talking about, long press a verse and then tap the globe).

> When playing recitation, the translation moves with the currently recited verse.

we already do this in the full screen view - if not then it'll be fixed as part of #709 insha'Allah.

> Translation area has configurable height.

this one goes hand in hand with the first point.

please let me know if i missed anything or if you have any suggestions,
walsalam 3alaikum.

ozbek:
> Moreover, the translation area itself shows all the verses of the page and is scrollable.

That's supported, albeit not in scrolling mode. You can use previous/next buttons at the bottom of translation area to go to other ayahs.

amynbe:
السلام عليكم
Thanks for agreeing with points 1 & 3.

Regarding point 2, i.e scrollable translation area vs previous/next buttons, the problem is: since the translation is often in a left--to-right language while Arabic is RTL, I personally get easily confused which button I have to tap.

Moreover, I find it more helpful for human memory/understanding to have 2 verses scrollable/visible at once than having to click to switch to another verse and hide the current one (verses are sometimes the same sentence split in 2 verses). I acknowledge this is possible in the full-screen view, but honestly I don't see much advantage having a separate translation view if the integrated one is properly usable. And as a learner, I'd rather not use the separate view because of the different Arabic text layout (with Ayat I never feel the need of it -- Ayat has different usability issues though).

Thanks again and I hope to read you soon.

PR #751, opened 2016-12-31
Translations

ahmedre:
This is a (mostly functional) work in progress - I am posting it here to get some feedback. It introduces a quran.com style translation view in place of the existing one.

in addition, it switches to `RecyclerView` for translations.

things left to do:
- [x] improve night mode (the ayah box in particular is quite bright, as are the lines, etc)
- [x] remove the last divider before a sura header
- [x] fix some bad performance that i introduced in this patch (translation presenter, replace the ayah view with a custom view).
- [x] re-implement highlighting of ayahs
- [x] the long press selection of translation and ayah text at the same time is not possible (they used to be one `TextView`, now they are 2), so need to consider replacing this functionality. *update* - for now, marking as done since translations can be long pressed and selected, as can the Arabic. in the future, we can consider something better for sharing.
- [x] [new feature] support multiple translations *update* - this is done, just needs ui to expose it

screenshots:

phone:

screenshot_1483174380

tablet:

screenshot_1483175467

the old implementation, for comparison:

screenshot_1483176088


ahmedre:
cc @ozbek @hosainnet

ahmedre:
here's a screenshot of night mode - note that the brightness of the text is a setting that can be overridden (and that also affects most of the text and dividers here).

screenshot_1483256706

app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBAdapter.java
```diff
@@ -1,73 +1,63 @@
...
   }
 
+  @NonNull
   public List<LocalTranslation> getTranslations() {
-    if (mDb == null) {
-      open();
-      if (mDb == null) {
-        return null;
-      }
+    List<LocalTranslation> cached = cachedTranslations;
```
ozbek: What's the `cached` variable for?

app/src/main/res/layout/quran_translation_arabic_row.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+          android:id="@+id/text"
```
ozbek: This does not seem to follow current spacing style.

app/src/main/res/layout/quran_translation_text_row.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+          android:id="@+id/text"
```
ozbek: Ditto.

app/src/main/res/layout/quran_translation_translator_row.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+          android:id="@+id/text"
```
ozbek: Ditto.

ahmedre:
re: the sura/verse numbers - yep, will do insha'Allah. part of me thinks there shouldn't be this text brightness setting and it should be replaced with 3 default settings - light, medium, and dark or something.

re: code indentation - not sure, the style is setup in Android Studio and i re-indented and there's no change there. i think xml rules are different than java in that it indents under the first attribute.

ozbek:
If you get `xmlns:android="http://schemas.android.com/apk/res/android"` to the next line, AS should align it correctly.

app/src/main/java/com/quran/labs/androidquran/database/TranslationsDBAdapter.java
ahmedre: i'll add a comment - but because the value of `cachedTranslations` can be replaced by a different thread going through `writeTranslationUpdates`, we keep a reference in `cached` (since `cachedTranslations` is volatile, this will be the latest value) and return it (otherwise, between the check and return, the value could change - we want any valid list of local translations, and it's okay if we get a slightly stale one here).

hosainnet:
I switched to the branch but can't see the new view, how do I trigger it?

ahmedre:
open the quran screen, tap the screen to see the Toolbar, and then tap the globe

app/src/main/java/com/quran/labs/androidquran/data/QuranDataProvider.java
```diff
@@ -66,15 +69,26 @@ private static UriMatcher buildUriMatcher() {
...
   }
 
   @Override
   public Cursor query(@NonNull Uri uri, String[] projection, String selection,
       String[] selectionArgs, String sortOrder) {
+    Context context = getContext();
+    if (!didInject) {
+      Context appContext = context == null ? null : context.getApplicationContext();
+      if (appContext instanceof QuranApplication) {
+        ((QuranApplication) appContext).getApplicationComponent().inject(this);
```
hosainnet: Would this work in onCreate instead? maybe then we can remove `didInject`.

ahmedre: intentionally did this, because from the documentation for `ContentProvider`:

> You should defer nontrivial initialization (such as opening, upgrading, and scanning databases) until the content provider is used (via query(Uri, String[], String, String[], String), insert(Uri, ContentValues), etc).

ahmedre:
screenshots with highlighting:

normal mode:

screenshot_1483318801

night mode:

screenshot_1483318810

ozbek:
It seems to me the repetition of translator's name is a little disturbing.  Does it have any (planned) purpose other than having resemblance to quran.com?

ahmedre:
yes, the intention is that this patch will support multiple translations, at which point, you need the translator's name before each one.

mohamede1945:
I might agree with @ozbek here too for the repetition of translator's name. What do you think of not repeating the translation source if there is only one translation selected but repeat it if multiple translations selected? In my opinion, a high percentage of users will just use one translation.

ATouhou:
@mohamede1945 Your assumption can be verified with Google Analytics, if we track events such as number of translations each session has activated  @mmahalwy

mmahalwy:
@mohamede1945 @ATouhou +1 on that. Definitely should be able to look into that data or gather it. I don't think at the moment we do track that, although we used to back in the day.

app/src/main/java/com/quran/labs/androidquran/common/LocalTranslation.java
```diff
@@ -19,4 +19,16 @@ public LocalTranslation(int id, String filename, String name,
     this.url = url;
     this.version = version;
   }
+
+  public String getTranslatorName() {
+    final String result;
+    if (this.translatorForeign != null) {
+      result = this.translator;
```
ozbek: You probably wanted to put `result = this.translatorForeign;`?

app/src/main/java/com/quran/labs/androidquran/common/LocalTranslation.java
```diff
@@ -19,4 +19,16 @@ public LocalTranslation(int id, String filename, String name,
     this.url = url;
     this.version = version;
   }
+
+  public String getTranslatorName() {
+    final String result;
```
ozbek: Is this variable necessary? Can't just return `this.x`?

app/src/main/java/com/quran/labs/androidquran/common/LocalTranslation.java
ahmedre: jazakAllah khairan will fix

app/src/main/java/com/quran/labs/androidquran/common/LocalTranslation.java
ahmedre: could, just preferring a single return

hulk9914:
السلام عليكم,
In this issue #722, It was required the following

>    Translation covers text but you can still scroll up to show the verses at the bottom of the pages.

I have an idea for solving this issue on tablets.
The idea is utilizing the TabletView for showing the quran page where an aya is highlighted in one side and the translation panel in the opposite page. We can also force the TabletView if the user is reading in portrait mode and required to show the translation of a chosen aya.

The benefit is the screen will be splitted horizontally into two separate pages (one for quran page and the other for translation panel) instead of splitting both quran pages vertically and lose half of them  both.
و جزاكم الله خيرا

ahmedre:
wa3laikum alsalam,
i think this is a good idea - i kept this in mind while doing a lot of the tablet refactoring i did in the past few patch sets, so this is now a lot easier than it used to be. i'll open a task insha'Allah.
walsalam 3alaikum.

ahmedre:
going to go ahead and merge this since things mostly work. there are a few improvements that are needed, insha'Allah will file tickets and/or fix these.

ozbek:
JazakAllah khair, awesome work.

ahmedre:
waiyyakum, and jazakAllah khairan for your help and reviews!

PR #752, opened 2017-01-02
Ayah number: use same text brightness as the translation text in night mode

ozbek:
I am not entirely sure if this is the correct approach. So, please criticize abundantly :)


ahmedre:
lgtm, jazakAllah khairan

PR #761, opened 2017-01-09
Fix build error

ozbek:
Error log:

> Error:Conflict with dependency 'com.android.support:support-annotations' in project ':app'. Resolved versions for app (25.1.0) and test app (23.1.1) differ. See http://g.co/androidstudio/app-test-app-conflict for details.

Tested only with Madani debug build, did not test with tests.


ahmedre:
jazakAllah khairan

PR #763, opened 2017-01-09
Update toolbar in translation view

### LINKED ISSUE #757: Toolbar doesn't update when going to translation view
when the last view visited is translations, we correctly go to the translations view - however, the Toolbar doesn't reflect this and is the Toolbar without the spinner (i.e. the reading mode).

cc @ozbek

ozbek:
Apparently, `translationItems` will not have enough time to get populated in `onCreate()` when we try to update the toolbar with translation items. So, let's call `updateActionBarSpinner()` once we get `translationItems` updated.

Fixes #757.


ahmedre:
jazakAllah khairan

Issue #750, opened 2016-12-30
Introduce functional/UI testing


hosainnet:
Salam. 

As Android uses a lot of final/static classes and methods, unit testing can be a little difficult and it would be good to have some more test coverage in the form of functional/ui tests (testing against a running app)

Some things to consider:

- What framework to use? e.g [Espresso](https://developer.android.com/training/testing/ui-testing/espresso-testing.html)?
- How will tests run on CI
- All external HTTP calls should be mocked  (something like [this](https://github.com/Confluex/confluex-mock-http) could work well)
- Use a decent test runner ([Spoon comes to mind](https://github.com/square/spoon))

Maybe for the purposes of this task, we can setup the framework and pick just a single activity to try an test it first, then use it going forwards.

What is everyone's thought on this?

ahmedre:
wa3laikum alsalam -
definitely - i've been rewriting a lot of code to try to split out ui logic and make things easier to test. do you want to take this?

- +1 for Espresso
- CI - Travis should support it, but if not, we can switch to CircleCI since they also have free plans for open source projects that we can use.
- Mocking HTTP - i've used OkHttp's MockServer to test [TranslationManagerPresenter](https://github.com/quran/quran_android/blob/master/app/src/test/java/com/quran/labs/androidquran/presenter/translation/TranslationManagerPresenterTest.java) and it works pretty well without adding any new dependencies.
- no problem using Spoon, but do we need it? i thought part of its use was running tests on multiple devices and capturing screenshots?

walsalam 3alaikum

hosainnet:
OK great, yes I'll take it inshallah.

We might need to think about how we structure our tests, e.g if we want to put Espresso tests in a different package etc.

Re Spoon, you're right it's not that necessary but it generates a neat HTML report and can take screenshots which is handy if we're running tests headless. I'll leave it out for now but we can add it later if needed.

PR #764, opened 2017-01-09
Setup Travis CI

hosainnet:
Follow up to #750 

- Setup Travis CI to run Espresso tests (max Android version supported is 22)
- Added build status on the README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ android:
       - platform-tools
       - build-tools-25.0.0
       - android-25
-      - android-22
+      - android-22 # required for emulator below
       - extra-android-m2repository
       - sys-img-armeabi-v7a-android-22
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,17 @@ android:
       - platform-tools
       - build-tools-25.0.0
       - android-25
+      - android-22
       - extra-android-m2repository
+      - sys-img-armeabi-v7a-android-22
+
+before_script:
+  # Create and start emulator
+  - echo no | android create avd --force -n test -t android-22 --abi armeabi-v7a
+  - emulator -avd test -no-skin -no-audio -no-window &
+  - android-wait-for-emulator
+  - adb shell input keyevent 82 &
 
 script:
    - ./gradlew clean testMadaniRelease -PdisableCrashlytics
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,7 +9,17 @@ android:
       - platform-tools
       - build-tools-25.0.0
       - android-25
+      - android-22
```
ozbek: Can you add 

> Android 22 is the latest in Travis at this time: https://docs.travis-ci.com/user/languages/android/

comment around here so that it will serve as a reminder for the next person looking at here?

Also, would not this make the one previous line obsolete? Having to similar lines one after another looks suspicious to me (I am not familiar with Travis build).

ahmedre:
lgtm with the exception of @ozbek's point, which i agree with - can you just remove `- android-22`?

hosainnet:
I tried first without including `- android-22` but it's required because that's the latest supported version to create the emulator image here `- echo no | android create avd --force -n test -t android-22 --abi armeabi-v7a`

I left `- android-25` because we use `- build-tools-25.0.0` so I assumed it's required, let me try removing `- build-tools-25.0.0` and see if it still works

hosainnet:
Okay so removing `android-25` breaks the build with `Before building your project, you need to accept the license agreements and complete the installation of the missing components using the Android Studio SDK Manager.`

Shall I put it back? I've left a comment near v22

ozbek:
Probably, this will help with the license error: https://docs.travis-ci.com/user/languages/android/#Dealing-with-Licenses ?

[Apparently](https://docs.travis-ci.com/user/languages/android/#Pre-installed-components), it is OK to list several versions of android-XX components.

> To ensure the stability of your build environment, we recommend that you explicitly specify the required components for your project

Sorry for not being more straightforward in my previous comment. The note regarding my not familiarity with Travis CI was to warn to take my comment with caution. Sorry again for the unnecessary issues caused ):

hosainnet:
No worries brother, I'm new to Travis too haha :)

I think I was wrong, maybe we can revert to just using v25, v22 just happens to be pre-installed.

Let me try a couple of things on another branch and see how it goes so I don't pollute this one!

ahmedre:
i think Android 25 is needed because we compile against Android N (api 25 - android 7.1).

PR #766, opened 2017-01-12
Add .nomedia file to audio folders

### LINKED ISSUE #565: add .nomedia file to audio
i am on the fence about this one, but saw several complaints recently about it. i personally don't use my quran android audio as the audio on my device.

we should probably do this anyway for gapped audio.

any thoughts on whether or not we should also do this for gapless?

ozbek:
Fixes #565

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
+++ b/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
@@ -159,8 +159,8 @@ public class QuranFileUtils {
     return bitmap == null ? new Response(Response.ERROR_FILE_NOT_FOUND) : new Response(bitmap);
   }
 
-  private static boolean writeNoMediaFile(Context context) {
-    File f = new File(getQuranImagesDirectory(context) + "/.nomedia");
+  public static boolean writeNoMediaFile(String parentDir) {
+    File f = new File(parentDir + "/.nomedia");
     if (f.exists()) {
       return true;
     }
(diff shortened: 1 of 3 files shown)


ahmedre:
jazakAllah khairan - why not just put the code and add it within `AudioUtils.getLocalQariUrl()` (and `AudioManagerActivity`'s `download` method) - i think those are the only two places you'd need them? ideally there should be just one point - maybe change the latter to use `AudioUtils.getLocalQariUrl` and put the modification there?

ozbek:
Yeap, I will look into it.

app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
```diff
@@ -159,8 +159,8 @@ public static Response getImageFromSD(Context context, String widthParam, String
     return bitmap == null ? new Response(Response.ERROR_FILE_NOT_FOUND) : new Response(bitmap);
   }
 
-  private static boolean writeNoMediaFile(Context context) {
-    File f = new File(getQuranImagesDirectory(context) + "/.nomedia");
+  public static boolean writeNoMediaFile(String parentDir) {
```
ozbek: @ahmedre: are you OK with this change? I'd like to re-use this method.

ahmedre: yep that's cool with me

ozbek:
>put the code and add it within `AudioUtils.getLocalQariUrl()`

It appears, at the point when `getLocalQariUrl()` is called for the first time, the directory structure may not have been populated yet. Hence, `QuranFileUtils.writeNoMediaFile()` fails to create the file (because the parent folder does not exist yet).

>(and `AudioManagerActivity`'s download method)

I could not figure out where in `AudioManagerActivity` to put the code.  It seems to use `AudioUtils.getQariUrl()`, but that method has same above issue as `getLocalQariUrl()`.

ozbek:
Please use #776.

PR #767, opened 2017-01-13
Remove QuranAyah

ahmedre:
QuranAyah was being used in many cases where SuraAyah should be used
instead. This patch replaces all those instances and removes QuranAyah.


ahmedre:
merging because it works and Travis is just timing out while running the ui tests.

PR #769, opened 2017-01-15
Added swipe-to-refresh in the translation manager view

### LINKED ISSUE #589: improve translation downloading screen
pull to refresh would be nice here since some people get empty cache files and can't get any data for ~1 week or without removing the cache file.

hosainnet:
@ahmedre I left the package refactoring out of this for now

Fixes #589


ahmedre:
jazakAllah khairan

PR #771, opened 2017-01-16
Update bookmarks & tags contextual action bar

habibr:
I just add several translation lines for the bookmarks & tags bar which have not been added to the indonesian strings.

This might resolve the "weird translation" that pop up in the Bookmark bar for the recent pages.


ahmedre:
jazakAllah khairan

PR #772, opened 2017-01-16
Create sura_names.xml

habibr:
Preparing for indonesian surah names translation.

### DIFF
--- a/app/src/main/res/values-in/sura_names.xml
+++ b/app/src/main/res/values-in/sura_names.xml
@@ -120,119 +120,119 @@
   <!-- If sura name has no translation, please add empty double quotes to the
        respective item body to keep Android Studio happy -->
   <string-array name="sura_names_translation">
-    <item>The Opener</item>
-    <item>The Cow</item>
-    <item>Family of Imran</item>
-    <item>The Women</item>
-    <item>The Table Spread</item>
-    <item>The Cattle</item>
-    <item>The Heights</item>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-in/sura_names.xml
```diff
@@ -0,0 +1,238 @@
...
+    <item>An-Nashr</item>
+    <item>Al-Lahab</item>
+    <item>Al-Ikhlas</item>
+    <item>Al-Falaq</item>
+    <item>An-Nas</item>
+  </string-array>
+
+  <!-- If sura name has no translation, please add empty double quotes to the
+       respective item body to keep Android Studio happy -->
+  <string-array name="sura_names_translation">
```
ozbek: You do not seem to be translating sura names into Indonesian at all. Please remove this array.

habibr: No, I have not. Sorry, it was too quick to press the pull request button.

ahmedre:
`sura_names_translation` looks good, but can you remove `sura_names` since they're the exact same as the ones in `values` so no need to duplicate?

ozbek:
`sura_names` looked the same to me also when I first saw them, but not all of them are the same :)

ahmedre:
ah good catch, jazakAllah khairan merging this

habibr:
Yes, the english and indonesian surah names are similar but different.

PR #775, opened 2017-01-17
Added ripple effect on translation manager rows

### LINKED ISSUE #758: press state on TranslationManager items
TranslationManager's items don't have ripples on them when tapped on 5x (nor a normal press state on 4x), we should fix this.

cc @hosainnet

hosainnet:
Fixes #758


ahmedre:
jazakAllah khairan - technically, i think all our ripples are wrong because they're all backgrounds - i think ripples should be in the foreground.

PR #776, opened 2017-01-18
Add .nomedia file to audio folder

### LINKED ISSUE #565: add .nomedia file to audio

ozbek:
Fixes #565.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
+++ b/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
@@ -159,8 +159,8 @@ public class QuranFileUtils {
     return bitmap == null ? new Response(Response.ERROR_FILE_NOT_FOUND) : new Response(bitmap);
   }
 
-  private static boolean writeNoMediaFile(Context context) {
-    File f = new File(getQuranImagesDirectory(context) + "/.nomedia");
+  private static boolean writeNoMediaFile(String parentDir) {
+    File f = new File(parentDir + "/.nomedia");
     if (f.exists()) {
       return true;
     }
@@ -180,9 +180,9 @@ public class QuranFileUtils {
 
     File directory = new File(path);
     if (directory.exists() && directory.isDirectory()) {
-      return writeNoMediaFile(context);
+      return writeNoMediaFile(path);
     } else {
-      return directory.mkdirs() && writeNoMediaFile(context);
+      return directory.mkdirs() && writeNoMediaFile(path);
     }
   }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
```diff
@@ -355,8 +355,21 @@ public static String getQuranAyahDatabaseDirectory(Context context) {
...
   public static String getQuranAudioDirectory(Context context){
-    String s = QuranFileUtils.getQuranBaseDirectory(context);
-    return (s == null)? null : s + AUDIO_DIRECTORY + File.separator;
+    String path = getQuranBaseDirectory(context);
+    if (path == null) {
+      return null;
+    }
+    path += AUDIO_DIRECTORY;
+    File dir = new File(path);
+    if (dir.exists() && !dir.isDirectory() && !dir.delete()) {
```
ozbek: This handles the case when there is an file (not a directory) with name `audio` under `quran_android`. Very unlikely case. If you think that is an overkill, please let me know and I will remove this block.

ahmedre: yes, let's not worry about it since the only way is if someone manually did that manually, and if they did, it's not our fault.

on the same note though, just so i am 100% sure - this writes `.nomedia` in `audio/` - will the various media players know that if there's `.nomedia` in audio, that there doesn't need to be one in `audio/{qari}`?

ozbek: >on the same note though, just so i am 100% sure - this writes `.nomedia` in `audio/` 

Yes, that's right; only under `audio/`.

> will the various media players know that if there's `.nomedia` in audio, that there doesn't need to be one in `audio/{qari}`?

If the media player is written correctly and gets data from framework's mediastore, then `.nomedia` in the root folder is enough. But I am not sure regarding the old Android versions. I will test on an emulator with older platform later and report here...

ahmedre: sounds good, jazakAllah khairan - ping me when it's ready to merge

ozbek: + works as expected on emulator with API level 14. Ready to be merged.

ahmedre:
jazakAllah khairan

PR #778, opened 2017-01-26
Experimentally show images for translations

ahmedre:
This patch implements per ayah images for translations by cutting pieces
of the existing images and using those to draw the images.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/common/AyahBounds.java
+++ b/app/src/main/java/com/quran/labs/androidquran/common/AyahBounds.java
@@ -1,5 +1,6 @@
 package com.quran.labs.androidquran.common;
 
+import android.graphics.Rect;
 import android.graphics.RectF;
 
 public class AyahBounds {
@@ -19,6 +20,10 @@ public class AyahBounds {
 
   public RectF getBounds() {
     return new RectF(bounds);
+  }
+
+  public Rect getBoundsAsRect() {
+    return new Rect((int) bounds.left, (int) bounds.top, (int) bounds.right, (int) bounds.bottom);
   }
 
   public int getLine() {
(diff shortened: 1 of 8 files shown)


ahmedre:
curious to get some feedback on this and really am not sure whether i should merge this or not.

so what it does is it displays images for each ayah in translation view. it does this by cutting out the relevant bounds of the image and displaying them.

in a good case, it looks like:
screenshot_1485394406

(the spacing between the first and second rows and the third and last rows is something that can be fixed).

there are a few downsides to this approach:

1. it is strongly dependent on our ayahinfo bounds being exactly right. for example, consider this:
screenshot_1485394456

when we cut this out, we end up with:
screenshot_1485394468

if you look closely, you'll notice the kasra is missing from the first letter of the first ayah. the reason is because of the first screenshot (the bounds are not correct, they ignore the first letter).

2. it doesn't properly support wrapping - ex:
screenshot_1485395266

so in these cases, we end up with things on 2 lines in a way that doesn't flow nicely. we could support these specific cases by merging them in the lines (we do merge now, but only when the overall width is less than the screen width), but there will be other cases where we get non-ideal views similar to these.

3. when we merge lines that were split in the page view, they sometimes appear unreasonably close together.
screenshot_1485395405

for example, ayah 4, the first word in the full page view is on one line, and the rest of the ayah is on the other line. when we put them besides each other, it looks too close. we can doctor this by adding some padding whenever we join lines, but if the font already has padding, it could have the reverse effect.

thoughts?

ozbek:
Why do you want to use images?

ahmedre:
people in the past have complained because the tashkeel is written slightly differently in the written text than it is in the uthmani drawn version - so there are things they think are mistakes and i have to explain that, no, it's not a mistake, it's written this way when it's written using imlaei fonts.

so i figured to get rid of this issue, we could just use images for each sura/ayah, but i wanted to avoid having to force a download if possible.

i am thinking to abandon this for now because it's not great as is and it will be very difficult to fix all the highlighting bounds to be perfect.

ozbek:
Oh well, I did not mean to entirely close it though :)

ahmedre:
lol i know, but writing my reply made me think, "if people are getting upset about things that are correct because they think they're wrong, what about things that are actually wrong, like parts of the tashkeel missing or cut off?" - as much as i love the solution, i guess it's a no-go and we'd have to either download fonts and render (or images).

PR #779, opened 2017-01-26
Use an observable instead of a subject

ahmedre:


ahmedre:
not actually going to merge this, just showing an alternative way this could be done. a third alternative is using a [relay](https://github.com/JakeWharton/RxRelay/)

PR #780, opened 2017-01-27
Use `external-files-path` instead of `external-path`

ozbek:
Because, we are actually using `external-files-path` albeit in hard-coded way. This should now fix exporting data in all other build variants and flavors as well.


ozbek:
Did not test, but prior to this change export in Qaloon Debug should fail.

ahmedre:
nice, jazakAllah khairan!

Issue #220, opened 2012-12-11
bookmark translation and tafseer


Somaiagabr:
add a way to bookmark from within the translation view

PR #797, opened 2017-02-02
Add an easier way to copy an ayah and translation

ahmedre:


ahmedre:
this is still a work in progress - it's not complete yet.

ozbek:
Hmm... While you are at it, can you consider issue #220 as well: add a context menu with share and bookmark options?

ahmedre:
@ozbek great point masha'Allah - let me think this through some

ahmedre:
abandoning in favor of 2bbfa7c

ahmedre:
btw, this patch makes #220 pretty straight forward to do (add an option for bookmarks in the `share_menu.xml` (and rename it, this is only used by translations), and then make sure that `TranslationPresenter` calls `toggleBookmark` (which is currently `private`)

Issue #799, opened 2017-02-04
Suggestion: jump to ayah by using EditText instead of drop-down Spinner


fikr4n:
السلام عليكم

Is there any planning to change drop-down Spinner of ayah number in "jump to page or surah & ayah" dialog to text field? I think it is easier to type ayah number instead of scrolling in case that the surah is long (has many ayahs). As for the surahs drop-down, how about replacing it with auto-complete text field which also has drop-down (like in the search box at the top bar)?

It's just an idea, based on my experience of a little difficulty when navigating.

fikr4n:
What I meant was like [this](https://github.com/fikr4n/quran_android/tree/text_jump). It is only a prototype, of course the code is still dirty.

ozbek:
Can you clean up the change (eg., remove `android.util.Log`, commented out code) and send a pull request? It'd be easier to follow your changes that way.

ahmedre:
i think it should be both autocomplete and spinner

fikr4n:
I've created the [pull request](https://github.com/quran/quran_android/pull/805). Not sure if it is what @ahmedre called "both autocomplete and spinner".

ahmedre:
jazakAllah khairan

PR #805, opened 2017-02-11
Jump to surah and ayah by allowing user to type (issue #799)

fikr4n:
This is the proposal that we've discussed in issue #799.

I want to note that:

1. I assume that all Android devices' numeric soft keyboard use Western Arabic digits (123...) instead of Eastern Arabic digits (١٢٣...), so that ayah number uses Western Arabic digits.
2. When testing by using Arabic, typing "ا-ل-أ-ع" does not make "الأعراف" to show up. It turns out that the sura names in string resource are not normalized, so I've a suggestion to normalize them, probably by using [NFKC](http://unicode.org/reports/tr15/#Norm_Forms). Other issue is that some sura such as الإنفطار is written as `الانفطار`, even comparing by using collator with `PRIMARY` strength will not treat them as equal string.
3. There is a method and a class which can be extracted out of `JumpFragment`, but I am not really sure.
4. I've not renamed the variables and layout IDs yet, to make minimum changes for this prototype.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java
@@ -8,6 +8,7 @@ import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.LayoutRes;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.v4.app.DialogFragment;
 import android.support.v7.app.AlertDialog;
 import android.text.Editable;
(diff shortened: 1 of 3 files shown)


ahmedre:
just tried this - this is awesome masha'Allah, jazakAllah khairan!

ahmedre:
oh, and sorry for the delayed reply on this PR - sorry things have been busy - would love it if you would contribute more also insha'Allah.

fikr4n:
وجزاك الله خيرا
It's my pleasure that you allow me to contribute to this project.

I've just remembered that I've not changed the variable names yet, e.g. `ayahSpinner` is still `ayahSpinner` while it should be changed to `ayahText` or `ayahInput` since it's no longer a spinner :D.

ozbek:
@fikr4n, please consider closing #799 :)

ahmedre:
please feel free to open another PR for the variable renames :)

PR #818, opened 2017-03-17
Do not show duplicate translation upgrade dialog

ozbek:
When there is an update, "Translations update available" dialog is shown two times. This checks if the dialog is already being displayed.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/QuranActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/QuranActivity.java
@@ -286,6 +286,9 @@ public class QuranActivity extends QuranActionBarActivity
   }
 
   private void showTranslationsUpgradeDialog() {
+    if (upgradeDialog != null && upgradeDialog.isShowing()) {
+      return;
+    }
     showedTranslationUpgradeDialog = true;
     AlertDialog.Builder builder = new AlertDialog.Builder(this);
     builder.setMessage(R.string.translation_updates_available);


ozbek:
**Observation:** [`QuranActivity.updateTranslationsListAsNeeded()`](https://github.com/quran/quran_android/blob/master/app/src/main/java/com/quran/labs/androidquran/ui/QuranActivity.java#L274) calls `showTranslationsUpgradeDialog()` based on the settings value. At the same time [`QuranDataActivity.runListView()`](https://github.com/quran/quran_android/blob/master/app/src/main/java/com/quran/labs/androidquran/QuranDataActivity.java#L531) puts `EXTRA_SHOW_TRANSLATION_UPGRADE` value based on the same settings value. Hence, the duplicate dialog.

I was not sure why the the logic was coded that way, so did not know whether I should change it and remove one of the checks. @ahmedre or anyone else who has more insight on this, can you please make any suggestions?

ahmedre:
yeah i am not sure, no idea what i was thinking when i did that, but it definitely looks wrong. one idea is to remove the passing of the parameter from `QuranDataActivity` and to always call `updateTranslationsListAsNeeded`, which will then notify about the update the next time the app runs (alternatively, we can connect a callback from the presenter to know when translations are updated and then show the dialog).

please feel free to give a different PR properly fixing this or i can fix it, as you wish - jazakAllah khairan

ahmedre:
btw, as you pointed out, that's why the dialog can appear twice - `onCreate` calls `showTranslationsUpgradeDialog` because it gets the parameter from `QuranDataActivity`, and then at the end of `onCreate`, we call `updateTranslationsListAsNeeded`,  which again calls `showTranslationsUpgradeDialog`.

PR #822, opened 2017-03-20
Jump to ayah by typing text: rename variables

fikr4n:
Changes:

- Rename variables and add some comment/javadoc to avoid confusion of those who read this code
- Use lambda

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java
@@ -8,6 +8,7 @@ import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.annotation.LayoutRes;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java
```diff
@@ -132,50 +125,47 @@ public void afterTextChanged(Editable s) {
...
+      try {
+        dismiss();
+        String pageStr = pageInput.getText().toString();
+        if (TextUtils.isEmpty(pageStr)) {
+          pageStr = pageInput.getHint().toString();
+          int page = Integer.parseInt(pageStr);
+          int selectedSura = (int) suraInput.getTag();
+          int selectedAyah = (int) ayahInput.getTag();
+
+          if (activity instanceof QuranActivity) {
```
ahmedre: i know this is existing but we should really clean this up. implement an interface in both QuranActivity and PagerActivity and just be like if it's this interface call jumpAndHighlight

fikr4n:
وجزاك الله خيرا

PR #802, opened 2017-02-06
Add Translation activity to spinner

uzairmad:
Added link to translations activity from spinner per https://github.com/quran/quran_android/issues/760, please review when you have a chance. thanks

PR #824, opened 2017-03-20
Add shared inflation for translation spinner

uzairmad:
Fix to use same view for translation spinner cases - mentioned at the end of #802


ahmedre:
jazakAllah khairan

uzairmad:
wayyakum

Issue #809, opened 2017-02-22
translation names cut off in spinner


ahmedre:
this is via an email report i got in

![screenshot_2017-02-21-10-44-46](https://cloud.githubusercontent.com/assets/47771/23199691/cfb3bc64-f884-11e6-9adf-f2d2e053f07c.png)

ozbek:
Why the spinner is not using localized name?

ahmedre:
good question, i am not sure - maybe it has to do with when he downloaded the one that is showing in Arabic versus the others

ozbek:
This did not get closed automatically...

ahmedre:
jazakAllah khairan

PR #826, opened 2017-03-21
Fix issue #809

nyitgroup:
I tried to make a new clean branch and fix the spinner at the source.


ahmedre:
i don't think this is the right fix - i added the code in QuranSpinner to make sure the drop downs with the sura/ayah numbers properly have the correct size. do you know why the workaround code isn't working in this case?

PR #827, opened 2017-03-21
Fix issue #809 Quran spinner  layout width

nyitgroup:
This seems to provide the maximum width available in the title area. 
I disabled the code to calculate width.
I think it is only measuring the linear layout and not the views inside the layout.


ozbek:
Awesome one liner! I'd use `match_parent` for the sake of consistency with the existing code though...

PR #828, opened 2017-03-21
Fix issue #809 Quran spinner  layout width match parent

nyitgroup:
Using match parent and leaving code in Quran spinner unmodified.


ahmedre:
jazakAllah khairan!

nyitgroup:
wayyakum

Issue #806, opened 2017-02-16
add dual pages mode preference on all devices


ahmedre:
this checkbox is hidden on phones, but enough people asked for it that it may make sense to show the preference everywhere.

PR #830, opened 2017-03-22
Enable tablet mode for phones

Ahmed9914:
Issue #806

### DIFF
--- a/app/src/main/res/values/booleans.xml
+++ b/app/src/main/res/values/booleans.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
-  <bool name="is_tablet">false</bool>
+  <bool name="is_tablet">true</bool>
   <bool name="use_tablet_interface_by_default">false</bool>
   <bool name="show_surat_prefix">true</bool>
   <bool name="show_sura_names_translation">false</bool>
--- a/app/src/main/res/values/booleans.xml
+++ b/app/src/main/res/values/booleans.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
-  <bool name="is_tablet">true</bool>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/booleans.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
-  <bool name="is_tablet">false</bool>
+  <bool name="is_tablet">true</bool>
```
ozbek: Rather than setting this to true, I'd remove it, and all references and methods using it.

Ahmed9914: I'll work on that.

selmanon: @Ahmed9914 Do you mean that you will work on issue #806 ?

ahmedre:
jazakAllah khairan for doing this - i'll merge this, but can you make another patch to rename "isTablet" to something like "isDualPagesMode" or something?

Ahmed9914:
I think you mean to change `prefs_tablet_mode_title` from "Tablet Mode" to "Dual page mode" as the method `isTablet` is checking the screen width, so its name is suitable to its functionality.

Please clearify this point.

JazakAllah khairan

ahmedre:
i mean both - isTablet is only used for dual screen mode, so the variable name shouldn't contain anything about tablets in it anymore (neither should the strings).

ozbek:
@Ahmed9914, while you are at it, would you please also remove `bool name="is_tablet"` definitions from other `values-xxx` folders? Also, please note that there is already `QuranUtils.isDualPages()` method.

app/src/main/res/values/booleans.xml
ozbek: @selmanon, apparently, he meant to work on my above comment :)

Ahmed9914:
@selmanon sorry, I didn't notice your question. yes as @ozbek said , I meant to work on my above comment

PR #831, opened 2017-03-22
#790 - fixed removing headphones doesn't pause audio

### LINKED ISSUE #790: removing headphones doesn't pause audio
this is bad, we should pause audio if the headphones are disconnected. we actually have code for this, so there's some bug causing it to not be triggered (the whole "becoming noisy" bits).

hosainnet:
- Added listener for an ACTION_AUDIO_BECOMING_NOISY which is triggered when headphones are unplugged:
https://developer.android.com/reference/android/media/AudioManager.html#ACTION_AUDIO_BECOMING_NOISY

Sorry for the delay on this!

Fixes #790

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
@@ -2144,8 +2144,9 @@ public class PagerActivity extends QuranActionBarActivity implements
   private class NoisyAudioStreamReceiver extends BroadcastReceiver {
     @Override
     public void onReceive(Context context, Intent intent) {
-      if (AudioManager.ACTION_AUDIO_BECOMING_NOISY.equals(intent.getAction())) {
-        // pause audio when headphones are unplugged
+      if (AudioManager.ACTION_AUDIO_BECOMING_NOISY.equals(intent.getAction())
+          && audioStatusBar.getCurrentMode() == AudioStatusBar.PLAYING_MODE) {
+        // pause audio when headphones are unplugged and audio is playing
         onPausePressed();
       }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -2132,4 +2140,14 @@ public void onPanelExpanded(View panel) {
     public void onPanelAnchored(View panel) {
     }
   }
+
+  private class NoisyAudioStreamReceiver extends BroadcastReceiver {
+    @Override
+    public void onReceive(Context context, Intent intent) {
+      if (AudioManager.ACTION_AUDIO_BECOMING_NOISY.equals(intent.getAction())) {
```
hosainnet: Just spotted something, I should only do this if audio is playing

ozbek: Yeap, probably need to `registerReceiver()` when starting the playback.

hosainnet: @ozbek I just added an extra check to see if the current mode is playing before pausing.

I guess what you suggested could be more efficient, but I'll have to modify the unregister logic to ensure it only happens under certain conditions.

What do you think?

ozbek: Yes. 

`PagerActivity` is already cluttered with lots of logic. Also, IMHO, handling service logic (audio playback) in an activity is not a good idea. I did not take a close look, but perhaps, this could better fit to `AudioService` instead.

ahmedre:
i think wa Allahu a3lam the intention of `AudioFocusHelper.java` was to catch these and handle them. unfortunately, from what i can tell, it's not setup - i.e. we don't `register` anywhere - so i think if we register in `onCreate` of `AudioService` (and unregister in `onDestroy`).

in case of becoming noisy, just call `onLostAudioFocus(false)`.

ahmedre:
jazakAllah khairan!

Issue #458, opened 2014-10-27
widget


ahmedre:
from an email we got (and i think we should definitely think about this, these are some good ideas masha'Allah):

Assalamu Alaikum and jazakumumullahu katheeran

I would like to urge the team to not miss a great opportunity for much reward. Two features _in a widget_ form will do wonders for the ppl of Quran, the family of Allah, the best Muslims:
1. Direct Surah/Ayah jump: People learning islam and daa'ees especially find references to Aayaat on the move and need to cross check or quote it or bookmark for tafsir study. Readers of physical mushaf at home might want to continue at the mosque for short reading. A widget that takes input like (23:21) and then jumps or book marks or shares immediately will be super.
2. Arabic search jump: for same read on move or quote on move  as above - on the move you want to look up ayat...opening app takes time. Widget launches straight to translate/bookmark/share

So while typing this email i could quote ayaat very quickly, especially if you can design a keyboard add-on (that's third feature actually?)

Jkkk

Salam.

yFotouh:
salam 3alaikum,
i was wondering if my pull request reached you for the widget feature?
Wa jazakum allah khayran

ahmedre:
wa3laikum alsalam,
i just tried it today, i am really sorry, i forgot. i like the idea behind this and the fact that it provides quick access, but the ui needs to improve. i especially love the one touch to get to the last page you were reading, i think that's great!

how do you feel about us actually having a list of bookmarks in the widget? i can try to see if i can get us some designs (will take at least 2 weeks though insha'Allah) so we can iterate on this and get it shipped out?

walsalam 3alaikum.

yFotouh:
السلام عليكم,
Well glad you liked it , and as you said the designs needs to improve ,and in fact i have not spent time at all in modifying it since i'm not very good at designing  , and thought that anyone could just plug these icons with the suitable images for each of the 3 widget options or of coarse if someone would like to change completely how it looks.
I could work also on a fourth widget option to get bookmarks in the weekend , but still will actually not spend so much time on how it looks , but can do something basic design though like a custom list view with a quran icon somewhere in the list view plus of coarse the text reference of bookmark and again someone could just come out and plug the suitable image or maybe change it's position so it would look nice.
Regarding shipping, do you mean merging my branch with the master?
وجزاكم الله خيرا
@ahmedre

yFotouh:
@Hello :)
@ahmedre

ahmedre:
[...]

PR #838, opened 2017-04-02
Bookmarks Widget

nyitgroup:
issue #458 Add bookmarks widget. 
Bookmark shows sura name and page number. 
I tested it in English and Arabic 
UI layout colors, logo can be improved with feedback.


ahmedre:
salam 3alaikum,
jazakAllah khairan for working on this - i looked at this and at the search one and both work well and i think are a good step - but i am not really a fan of the design of either one.

i had mockups made sometime back on #458 - can you consider making the bookmarks widget looking like the second screenshot, and the first one (the search one) looking like the first screenshots (just with a search button that opens the dialog like you're doing now, and the second one opening the last page you were on, and maybe the text of Quran or the image launching the app?)

what do you think?
walsalam 3alaikum.

nyitgroup:
I will come by and both widgets as you suggested https://cloud.githubusercontent.com/assets/47771/7449070/1d90562e-f1e1-11e4-8d40-930702d0837d.jpg

PR #847, opened 2017-04-15
Improve Farsi translations

bateni:
There were a lot of mistakes, and many missing translations.  In particular, see the message in the first commit.


ahmedre:
jazakAllah khairan for doing this!

PR #851, opened 2017-04-16
Change tablet strings to dual-page in strings-fa to keep up with #846.

bateni:


bateni:
I have no idea why this pull request is failing on Travis CI.  It's a very simple change.  Basically #846 changed all instances of `prefs_tablet_mode_title`, etc. to `prefs_dual_page_mode_title`, etc.  However, #847 just added these keys in Farsi localization.  Therefore, #846 did not change the corresponding translations for Farsi.  Now we have three translated strings for Farsi that have no English default (and the settings menu shows the English defaults for dual-page keys).
@ahmedre @Ahmed9914

ahmedre:
it's unrelated - it's a timeout due to the instrumentation tests we added. jazakAllah khairan for doing this.

Issue #849, opened 2017-04-16
Audio panel does not show the sura names properly (in Farsi)


bateni:
This is assuming I have patched the updated Farsi translations: #848 #847  But it should occur without those, as well.

Unlike English and Arabic interfaces, sura names are not fully shown.  See the picture (and ignore the green-ish circles).
<img width="280" alt="screen shot 2017-04-15 at 9 22 18 pm" src="https://cloud.githubusercontent.com/assets/6666981/25067985/3faa1a7e-2222-11e7-91e1-3d693da20ced.png">

Then I thought maybe the problem is that there are no customized Farsi layouts.  I don't know what is the difference between layouts under `layout-ar` and `layout`, but in principle Farsi and Arabic layouts should be exactly the same (except for the localized strings).

So I copied Arabic layouts under `layout-fa` directory, etc.  This does not have the above patch for Farsi translations, but the result is similar.  See the picture.
<img width="279" alt="screen shot 2017-04-15 at 9 10 19 pm" src="https://cloud.githubusercontent.com/assets/6666981/25067999/b777d712-2222-11e7-92e8-6fe1e54bbcf5.png">

The spinner's width is good, and can theoretically fit the longer sura names.  But even a short name like "الرعد" is not fully displayed.

nyitgroup:
I can't replicate this problem on my phone on the release or development version. 
Which version of the App are you using?
Can you try the latest Dev version?

It might be related to this spinner issue?
https://github.com/quran/quran_android/commit/1de170c3bb9884ab5d6643880c662b6c71734c6e

bateni:
I can reproduce this with the latest version from github.  I just open the project in Android Studio and run the Debug binary on a simulator for Nexus 5, API 25 (with locale set to fa-IR).

<img width="301" alt="screen shot 2017-04-17 at 6 13 14 pm" src="https://cloud.githubusercontent.com/assets/6666981/25106867/e5535e36-2399-11e7-87a3-7d18869e90a7.png">

nyitgroup:
confirmed
![fati7ah_device-2017-04-18-200834](https://cloud.githubusercontent.com/assets/13177699/25158204/94fc5a2e-2474-11e7-9869-bde8ff9ef04b.png)

nyitgroup:
![fixed device-2017-04-18-210846](https://cloud.githubusercontent.com/assets/13177699/25159098/4b56f95e-247b-11e7-9448-1772bc497f1c.png)

I will upload potential fix soon

[...]

PR #856, opened 2017-04-19
FixAudioSuraName

nyitgroup:
Fix: Audio panel does not show the sura names properly (in Farsi) #849
Add weight 1 to linear layout child spinners to fix width

### DIFF
--- a/app/src/main/res/layout-ar-land-v17/audio_panel.xml
+++ b/app/src/main/res/layout-ar-land-v17/audio_panel.xml
@@ -57,11 +57,13 @@
               android:id="@+id/start_sura_spinner"
               android:layout_width="wrap_content"
               android:layout_height="wrap_content"
+              android:layout_weight="1"
               />
           <com.quran.labs.androidquran.widgets.QuranSpinner
               android:id="@+id/start_ayah_spinner"
               android:layout_width="wrap_content"
               android:layout_height="wrap_content"
+              android:layout_weight="1"
               />
         </LinearLayout>
         <LinearLayout
@@ -74,11 +76,13 @@
               android:id="@+id/end_sura_spinner"
               android:layout_width="wrap_content"
               android:layout_height="wrap_content"
+              android:layout_weight="1"
               />
           <com.quran.labs.androidquran.widgets.QuranSpinner
               android:id="@+id/end_ayah_spinner"
               android:layout_width="wrap_content"
               android:layout_height="wrap_content"
+              android:layout_weight="1"
               />
         </LinearLayout>
       </LinearLayout>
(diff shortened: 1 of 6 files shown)


bateni:
This seems to fix the problem but is sort of hacky...  why doesn't the spinner show the full name of the suras, although it has sufficient width for it?

Can you reproduce the problem for Arabic locale?  I can't.  Why doesn't it occur there?

nyitgroup:
I was looking for a general and quick solution. 
The spinner might be measuring width wrong for different languages. I will look into it, but I can't promise anything quick. :)

ahmedre:
yeah, this is pretty hacky and changes the desired behavior.

unfortunately, this is a pretty tricky problem. here are some notes:
1. if you look at the layout, you'll see that the text doesn't quite fill its parent (the `Spinner`), even though it should.
2. in `QuranSpinner`, we override `onMeasure` to correctly figure out the width of the dropdown - if you notice, the dropdown width is correct because of this. both `Spinner` and `AppCompatSpinner` calculate the dropdown width (and the spinner's width) based on the largest width of the first 15 elements, whereas `QuranSpinner` calculates the last 15 items, and compares that width to the width found from the first 15 items, and uses the largest of the 2.
3. if we disable all the logic in `QuranSpinner`, the  bug still happens, suggesting it's a bug either in `AppCompatSpinner`, or in one of the framework classes that it extends from (`Spinner` or `AbsSpinner`). my guess is that the bug is in `Spinner` itself.
4. if we ask the child to measure in `QuranSpinner`'s `onMeasure` once we decide on a new width to use, this gets overridden later on by `Spinner`'s `setUpChild` method, which measures the view using the `Spinner`'s own layout parameters as a basis.

the least hacky workaround i could come up with here is to explicitly update the spinner's layout parameters once we've determined the width. this will cause any further `setUpChild` calls to use the spinner's current layout parameters, which have a fixed width, and consequently, it solves the problem. i'll push a patch for this insha'Allah.

as for your question - "why does it work for English and Arabic, but not for Farsi?" i think the reason is that there is some bug in the text measurement logic, specifically when you combine Farsi numbers with Farsi text. here are 2 proofs for this:

1. remove the for loop at `AyahPlaybackFragment`, line 170, which prepends the localized number to the sura name - re-run the app and you'll see that it now works, just like Arabic and English does.
2. undo change 1 if you made it, and then jump into the `QuranUtils.getLocalizedNumber` method, and force `isArabicNames` to true, then re-run. you'll again see that it works just fine.

ahmedre:
i will close this, please re-open if you think it should be re-opened. jazakAllah khairan!

Issue #860, opened 2017-04-29
Incorrect comma in Juz' list


bateni:
In the Juz' list page, English comma is used for Farsi and Arabic.

<img width="290" alt="qa-comma-bug" src="https://cloud.githubusercontent.com/assets/6666981/25555893/bae5b3e0-2cbf-11e7-9152-4ff13376ba70.png">

ahmedre:
jazakAllah khairan for fixing

PR #861, opened 2017-04-29
Fix comma problem in the Juz' list page (for Farsi, Arabic, etc.).

bateni:
Currently the English comma is used, which is  incorrect.  #860 

To fix, I switched to use 'sura_ayah_notification_str' instead of building the string ad-hoc.


ahmedre:
jazakAllah khairan

PR #864, opened 2017-05-08
Set translatable="false" for referenced strings

ozbek:
These strings are reference to other existing ones, so they don't need translation in other locales.


ahmedre:
jazakAllah khairan

PR #865, opened 2017-05-08
Remove "plural_recent_pages" from values-in/strings.xml

ozbek:
There are two reasons for removal:
 1. `"plural_recent_pages"` is set as `translatable="false"` in `values/strings.xml`.
 2. `"one"` quantity is irrelevant in Indonesian, so it needs a proper revision.


ahmedre:
jazakAllah khairan

Issue #834, opened 2017-03-28
Translation view: jumping disables action bar spinner


ozbek:
1. Open the app in translation mode
2. Select 'Jump' menu and go to any page
3. Observe that action bar is no more a spinner and one can not (de-)select translations.

Ahmed9914:
@ozbek This is fixed now. Please close.

ozbek:
Uh huh? Why is this not auto-magically get closed? Probably, should have used the magic word (`Closes #834` or `Fixes #834`) in the PR :)

PR #877, opened 2017-05-23
Fix "Jumping disables action bar spinner in translation view"

Ahmed9914:
Issue https://github.com/quran/quran_android/issues/834

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
@@ -2006,6 +2006,9 @@ public class PagerActivity extends QuranActionBarActivity implements
         case R.id.cab_share_ayah_text:
           shareAyah(start, end, false);
           break;
+        case R.id.cab_share_ayah_image:
+          shareAyah(start, end, false);  // to be changed to shareAyah(start, end, false, image = true)
+          break;
         case R.id.cab_copy_ayah:
           shareAyah(start, end, true);
           break;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -1078,7 +1078,6 @@ private void updateActionBarTitle(int page) {
     String sura = QuranInfo.getSuraNameFromPage(this, page, true);
     ActionBar actionBar = getSupportActionBar();
     if (actionBar != null) {
-      translationsSpinner.setVisibility(View.GONE);
```
ozbek: There must be a reason why we are hiding translation spinner here (in fact, there is a separate method for updating actionbar spinner).

Ahmed9914: The problem is that `updateActionBarSpinner()` is calling `updateActionBarTitle()` . 
I tried to figure out the reason for hiding translation spinner, but I didn't notice the reason.
May be @ahmedre has any thoughts?

Ahmed9914: @ozbek You are right, may be the problem is in [this](https://github.com/quran/quran_android/blob/master/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java#L771) line, I'll close this PR for now and try to use `updateActionBarSpinner()` instead of `updateActionBarTitle()`.

PR #880, opened 2017-05-26
Fix "Jumping disables action bar spinner in translation view"

Ahmed9914:
This is a clean PR as the old one is messed up
Issue #834


ahmedre:
jazakAllah khairan

PR #882, opened 2017-05-27
Fix strange behaviour of Jump dialog on re-creation

fikr4n:
Issue:

- When user types sura name in Jump dialog (and completion/suggestion drop down displays) and then rotates the screen, it forces complete instead of displaying the drop down just like before.

Changes:

- Use a new approach instead of `onAttachedToWindow`.
- Create a specific event name instead of "borrowing" `onItemClick`.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
@@ -78,6 +78,7 @@ import com.quran.labs.androidquran.ui.fragment.TranslationFragment;
 import com.quran.labs.androidquran.ui.helpers.AyahSelectedListener;
 import com.quran.labs.androidquran.ui.helpers.AyahTracker;
 import com.quran.labs.androidquran.ui.helpers.HighlightType;
+import com.quran.labs.androidquran.ui.helpers.JumpDestination;
 import com.quran.labs.androidquran.ui.helpers.QuranDisplayHelper;
 import com.quran.labs.androidquran.ui.helpers.QuranPage;
 import com.quran.labs.androidquran.ui.helpers.QuranPageAdapter;
@@ -125,7 +126,8 @@ public class PagerActivity extends QuranActionBarActivity implements
     AudioStatusBar.AudioBarListener,
     DefaultDownloadReceiver.DownloadListener,
     TagBookmarkDialog.OnBookmarkTagsUpdateListener,
-    AyahSelectedListener {
+    AyahSelectedListener,
+    JumpDestination {
   private static final String AUDIO_DOWNLOAD_KEY = "AUDIO_DOWNLOAD_KEY";
   private static final String LAST_AUDIO_DL_REQUEST = "LAST_AUDIO_DL_REQUEST";
   private static final String LAST_READ_PAGE = "LAST_READ_PAGE";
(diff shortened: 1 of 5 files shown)


fikr4n:
I remembered @ozbek's suggestion to create interface instead of using the activity directly, so I also create a new interface. Not sure if it should be inner interface of `JumpFragment` or just like this.

fikr4n:
Brothers, I inspected the `QuranActivityTest` test case and found that it checks for the existence of "Surat Al-Fatihah" list item on the first page of `QuranActivity`. The problem is that the fragment scrolls to the latest visited sura. It makes the test case fails if the latest page is not Al-Fatihah. I think the test case should be updated so the scenario is identical for each time it is run/executed.

app/src/main/java/com/quran/labs/androidquran/widgets/ForceCompleteTextView.java
```diff
@@ -46,23 +52,39 @@ protected void onFocusChanged(boolean focused, int direction, Rect previouslyFoc
...
-   * use to force the completion, the listener will be called with position argument set to -1 and
-   * view argument set to null.
+   * Sets the listener that will be called to do force completion ({@link #onForceComplete(int,
+   * long)}).
+   */
+  public void setOnForceCompleteListener(@Nullable OnForceCompleteListener l) {
+    this.onForceCompleteListener = l;
+    post(() -> {
+      if (!isFocused()) {
+        onForceComplete(-2, AdapterView.INVALID_ROW_ID);
```
ozbek: Just curious, why did you use -2 here but -1 elsewhere for the first argument?

fikr4n: There is no real reason, just to make them different. I don't know if it will be useful in the future.

ozbek: Not sure about the future either, but it looks confusing **now** :)

fikr4n: Okay, I'll change it. Barakallah.

fikr4n: How about `AdapterView.INVALID_POSITION`?

ozbek: That sounds very reasonable.

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -799,12 +801,14 @@ public void onNewIntent(Intent intent) {
...
   }
 
+  @Override
   public void jumpTo(int page) {
     Intent i = new Intent(this, PagerActivity.class);
     i.putExtra("page", page);
     onNewIntent(i);
   }
 
+  @Override
```
hakimrie: since this method is override from interface, I think it is better to group it at the top class, below override methods from Activity.
Moreover this class should be re arranged so all public method at the top and all private method at the bottom 🤔

fikr4n: What do you think @ozbek?

ozbek: Grouping is a good idea, but at the top or bottom - that changes from person to person. I'd wait for @ahmedre on that.

ahmedre:
jazakAllah khairan, works better than the existing way.  i noticed on the emulator though that if the keyboard is visible and you switch orientations (from portrait to landscape) that the dropdown disappears - no additional completion seems to happen in this case unless you start erasing characters and re-typing.

for the `QuranActivityTest` issue, may you file a bug?

fikr4n:
Wajazakallahu khairan.

> for the QuranActivityTest issue, may you file a bug?

I've posted as #902.

PR #883, opened 2017-05-27
Normalize sura names (Arabic)

fikr4n:
The problem is that searching sura in Jump dialog by typing sometimes fails to match, it seems the problem is (1) the way the string compared and (2) sura names itself. In this commit I try to cover the problem (2) by normalizing the sura name by unicode NFKC and change some alef (ا) to alef with hamza below (إ). I create a separate PR instead of include it in #882 because I am not native to Arabic and not really sure how users in Arabic locale type sura name, so that I need your consideration.

### DIFF
--- a/app/src/main/res/values-ar/sura_names.xml
+++ b/app/src/main/res/values-ar/sura_names.xml
@@ -6,9 +6,9 @@
         <item>آل عمران</item>
         <item>النساء</item>
         <item>المائدة</item>
-        <item>اﻷنعام</item>
-        <item>اﻷعراف</item>
-        <item>اﻷنفال</item>
+        <item>الأنعام</item>
+        <item>الأعراف</item>
+        <item>الأنفال</item>
         <item>التوبة</item>
         <item>يونس</item>
         <item>هود</item>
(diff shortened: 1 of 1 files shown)


WailB:
I have faced this issue in my Quran app. The way I handled it is to provide two types of searches, approximation and exact. Exact is clear where the user have to enter the exact letters including whether hamza is above or below etc. While approximation ignores أ و ي and all their derivatives while maintaining the same position in the word as empty. So for example, the word مؤمنون will be searched as م من ن the spaces in between can be any letter as long the word consist of those letters in that order and the word is 6 letters long.

ahmedre:
not against adding `إ` where we don't have it yet (ex infitar, inshiqaq) but am curious as to how the others are different (i.e. some of the lines look exactly the same - why is the unicode character you replaced more normalized than the existing one?)

as @WailB said, we do something similar for search across the Arabic text - so there are two columns in the database, one for normalized text and one for actual text. normalized text has no tashkeel at all, whereas actual text does. i think (not sure if we do, if we don't, we should fix it) that we normalize the search string also and compare against the normalized text.

not against merging this, just would love more of an explanation on the question above - jazakAllah khairan!

fikr4n:
> some of the lines look exactly the same

They are visually identical and render exactly the same glyphs. However, the old strings are one character while the normalized ones are two characters. I tried typing ل and then أ, it entered as two characters so that they didn't matched the one character (unnormalized) ﻷ. However, I don't usually use Arabic in my phone (and I am not fluent in Arabic language), so that I leave the decision to those who use Arabic.

fikr4n:
> not against adding `إ` where we don't have it yet (ex infitar, inshiqaq)

Alhamdulillah it is not merged yet. I found it was my mistake as these two are actually *alef* (*hamza wasl*), it was a mistake to change them to *hamza under alef*. So I revert it.

PR #893, opened 2017-06-12
Improve highlighting logic for ayahs

ahmedre:
This makes highlighting for ayat look a lot better. It does this by
combining the highlight bounds to look like one consecutive block
instead of many disjoint lines.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/model/quran/CoordinatesModel.java
+++ b/app/src/main/java/com/quran/labs/androidquran/model/quran/CoordinatesModel.java
@@ -8,9 +8,13 @@ import com.quran.labs.androidquran.data.AyahInfoDatabaseHandler;
 import com.quran.labs.androidquran.data.AyahInfoDatabaseProvider;
 import com.quran.labs.androidquran.di.ActivityScope;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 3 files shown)


ahmedre:
before:
screenshot_1497255866

after:
screenshot_1497259866

this code works and handles all the cases i could think of, but i need to improve it (and need to write some tests for it) before merging it insha'Allah.

ahmedre:
cc @mohamede1945

mohamede1945:
Nice work, mashaa'Allah! I totally support it!

app/src/main/java/com/quran/labs/androidquran/model/quran/CoordinatesModel.java
```diff
@@ -46,6 +52,125 @@
...
+
+    AyahBounds middle = null;
+    for (int i = 1; i < (total - 1); i++) {
+      if (middle == null) {
+        middle = ayahBounds.get(i);
+      } else {
+        middle.engulf(ayahBounds.get(i));
+      }
+    }
+
```
hakimrie: this function is quite large, I think it is better to split it into more smaller function.
for example here we can directly
```
return middle != null
        ? normalizeAyahBoundsContainMiddleAyah(ayahBounds, middle)
        : normalizeAyahBoundsWithoutMiddleAyah(ayahBounds);
```

ahmedre: jazakAllah khairan, agreed - let me think about how to clean this up.

PR #911, opened 2017-10-06
Do not allow canceling storage permission rationale dialog

ozbek:
Canceling or dismissing (e.g., by touching outside the dialog) this dialog leaves the UI in unusable (blank) state.


ahmedre:
jazakAllah khairan

PR #927, opened 2018-01-12
Make QuranInfo usage non-static

ahmedre:
This patch makes the usages of `QuranInfo` (and, consequently, a number of other classes) non-static. This is an important step that insha'Allah paves the way for passing a data source dynamically and being able to eventually change the page type within the app.


ahmedre:
integration tests are still failing, will disable in another patch.

PR #946, opened 2018-04-16
Put ext definitions into single block

ozbek:
Also, bumped up some versions (dagger, OkHttp) and removed `buildToolsVersion` (recent android gradle versions automatically pick-up the most relevant version).


ahmedre:
jazakAllah khairan!

PR #955, opened 2018-06-02
Logo Change in order to create a more unified experience.

Riz-waan:
I have updated the logo to match the iOS and web version of the app. I believe this will create a unified experience. The logo change is based on the new Quran.com's logo and taken from the iOS version's logo.


faisalmohd83:
Any thoughts on this?

ozbek:
Current version is more appropriate than the proposed one. Perhaps, it's better for web and iOS versions to adapt Android one (:

DoozyDoz:
> Any thoughts on this?

@faisalmohd83 the current logo is used by the whole organisation (IOS app, Web too), why do you want it changed?

Issue #970, opened 2018-08-25
Wrong Juz in some specific case


Naheel-Azawy:
Alsalamu alykum,
Good news! an issue that is easy to fix! I just didn't have time to fix it on my own :(
This problem can be seen in sura Al-Maeda page 121. The 7th Juz indeed starts at that page at Aya 82 but most of this page is in the 6th Juz. In fact, if you check any printed Quran it can be seen that page 121 is marked under Juz 6.
I'm not sure if this problem occurs somewhere else and I'm not sure if this is a rule but I think that the right way of selecting the Juz for a page is selecting the Juz of the first Aya in the page.
Again I'm not sure but the problem can be probably fixed from com.quran.labs.androidquran.data.QuranInfo.getJuzFromPage()

And I almost forgot to thank you for your great work!

ahmedre:
wa3laikum alsalam -
yes, but the madani mus7af does show as @Naheel-Azawy said the previous juz' - so we should fix this in sha' Allah just to be consistent (also, other people have complained to me about this in the past as well so might as well fix it for consistency reasons).
walsalam 3alaikum.

ahmedre:
in addition to these 2, one complaint about -

- check also juz' 23 and juz' 26 to make sure they're correct
- 3/4 hezb 60 is one page late?

so let's double check these are correct also.

PR #971, opened 2018-08-25
fixed juz for pages 121 and 201

Naheel-Azawy:
This is in response to issue #970 that I've just opened.
Yes, I know it seems like a bad idea to hard code it but these are just two cases in the entire Quran.
I also preferred that because the data is correct, page 121 is indeed where Juz 7 starts and the same for page 201. It is just what is displayed on the page is what had to be changed.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/data/QuranInfo.java
+++ b/app/src/main/java/com/quran/labs/androidquran/data/QuranInfo.java
@@ -259,6 +259,11 @@ public class QuranInfo {
   }
 
   public int getJuzFromPage(int page) {
+    if (page == 121) {
+      return 6;
+    } else if (page == 201) {
+      return 10;
+    }
     for (int i = 0; i < juzPageStart.length; i++) {
       if (juzPageStart[i] > page) {
         return i;


ozbek:
These are just semantics, not as big deal as you have put it.

ahmedre:
i agree though that we should fix it because i checked the madani mushaf and as @Naheel-Azawy said, they show the previous juz' number at the top for those pages - so i want the app to be consistent with that.

i think this should be fixed in the underlying data structures though instead of in this particular way.

app/src/main/java/com/quran/labs/androidquran/data/QuranInfo.java
```diff
@@ -259,6 +259,11 @@ private String getSuraNameFromPage(Context context, int page) {
   }
 
   public int getJuzFromPage(int page) {
+    if (page == 121) {
+      return 6;
+    } else if (page == 201) {
+      return 10;
+    }
```
hakimrie: sorry, if it is required to be changed, why not change the data source directly (eg. in this file https://github.com/quran/quran_android/blob/master/common/data/src/main/java/com/quran/data/pageinfo/common/MadaniDataSource.kt) 🤔

Naheel-Azawy: Because the data is not wrong. Juz' 7, for example, indeed starts at page 121 but on the last aya in the page. The only thing that should be changed is when the user will be notified about the end of the Juz. In this example, Juz' 6 does not end in page 121 but in 122. If you write Juz' 7 on page 121, the user might think that Juz' 7 starts on the beginning of this page which is not the case.
Another scenario to be considered is if one would like to jump to the 7th Juz. In this case the app should go to page 121 and not 122 because it starts at the end of page 121.

Naheel-Azawy: Maybe a cleaner way of doing it would be having another array getPageForJuzArray2? with the two pages changed and adding a flag to the function to specify whether it should be the original one or the second one but I thought that this is too much for it?

hakimrie: personally I prefer there is another array.
I think current code has many hidden information that can't be easily understood just by reading the code

ahmedre:
jazakAllah khairan @Naheel-Azawy for doing this - i used your patch as the basis for #1019, which does what you're doing except that it does it in a different method (since, as you correctly pointed out, the juz' data itself is correct, it's only the display data that is different).

closing this as a result - jazakAllah khairan again!

Naheel-Azawy:
@ahmedre Sorry for disappearing for a while, I was really busy. And thanks for you great work!

PR #974, opened 2018-08-30
Update Android plugin and make buildtools explicit

ahmedre:


app/build.gradle
```diff
@@ -6,6 +6,7 @@ apply plugin: 'kotlin-kapt'
 
 android {
   compileSdkVersion deps.android.build.compileSdkVersion
+  buildToolsVersion '28.0.2'
```
ozbek: Why is this needed?

ahmedre: without it, it's defaulting to this version anyway (forced by the Android gradle plugin i think?) - so figured might as well set it explicitly so it's not resolving to something different without us knowing what it is.

ozbek: That's right, the tooling version is selected automatically depending on the Android Gradle version. [The prior change](https://github.com/quran/quran_android/pull/946) was intentional :)

ahmedre:
abandoning in favor of #978

PR #977, opened 2018-09-16
Audio cleanup

ahmedre:
This set of patches cleans up and improves the handling of audio from `PagerActivity`.


ahmedre:
I should note that I am working on a set of patches to continue this cleanup over to `AudioService` as well, after which point `LegacyAudioRequest` and its children should ideally go away.

PR #984, opened 2018-10-01
Define kotlin code style following kotlin code conventions

hakimrie:
Define kotlin code style following kotlin code conventions, this kotlin configuration borrowed from https://android.googlesource.com/platform/tools/metalava/+/master/.idea/codeStyles/Project.xml

PR #995, opened 2018-10-30
Convert class QariItem to Kotlin

hakimrie:
Please review & merged this PR https://github.com/quran/quran_android/pull/984 first to reduce changes

I will use this PR as base branch for other java to kotlin class conversion


app/src/main/java/com/quran/labs/androidquran/common/QariItem.kt
```diff
@@ -0,0 +1,21 @@
...
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+class QariItem(
+  val id: Int,
+  val name: String,
+  val url: String,
+  val path: String,
+  var databaseName: String? = null
```
ahmedre: will merge this, but please change this to be a `val` - the code in `init` doesn't add much value in this case.

ahmedre:
jazakAllah khairan!

PR #996, opened 2018-10-30
Convert class LocalTranslation to Kotlin

hakimrie:
please review PR #995 first before this PR


ahmedre:
jazakAllah khairan

ahmedre:
ideally i'd avoid replicating the getters, but this is a good excuse for us to move more code to Kotlin to clean that up in the future in sha' Allah :)

PR #1002, opened 2018-11-02
Force complete sura name before jumping (click OK)

fikr4n:
السلام عليكم

In Jump dialog, when moving from sura `AutoCompleteTextView` to other `TextView`, it's forced to complete because it loses focus. However when clicking OK, there's no unfocus event happens, so I propose to call it "manually".


app/src/main/java/com/quran/labs/androidquran/widgets/ForceCompleteTextView.java
```diff
@@ -34,6 +34,10 @@ private void init() {
     super.setOnItemClickListener((parent, view, position, id) -> onForceComplete(position, id));
   }
 
+  public void forceComplete() {
+    onForceComplete(AdapterView.INVALID_POSITION, AdapterView.INVALID_ROW_ID);
+  }
```
hakimrie: what if we convert this class to kotlin and set `AdapterView.INVALID_POSITION` and `AdapterView.INVALID_ROW_ID` as default value for `position` & `rowId` in function `onForceComplete`, so we no need to add this `forceComplete` function.

fikr4n: What do you thing @ozbek ?

ozbek: Converting to Kotlin aside, I don't think you can call `onForceComplete()` from `JumpFragment` class (at least, not without changing its scope definition).

ozbek: Yes (what I meant was converting to Kotlin needs to be done for other reasons:). What do you say @hakimrie?

hakimrie: I think all force complete should be called from Listener (OnForceCompleteListener) to prevent leaky abstraction.
it better to update `jump_dialog.xml` by adding this two attribute in parent layout:
```
android:focusableInTouchMode="true"
android:clickable="true"
```
to make `onFocusChanged` called when user click outside ForceCompleteTextView.
what do you think @fikr4n @ozbek ?

fikr4n: Not sure why, but I've tried to set those two properties in several views (trial and error) in `jump_dialog.xml`, but cannot achieve force completion when clicking OK button.

hakimrie: @fikr4n I am sorry for very slow response. You're right, It doesn't force completion if user click outside `jump_dialog.xml` layout, in only work if user click inside it. 

I propose we request parent layout focus programmatically before [dismissing](https://github.com/quran/quran_android/blob/3a6888406aff86ba5ada02c83f3881f028e66346/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java#L155) AlertDialog to make  sure`ForceAutoCompleteTextView.onFocusChanged` called and JumFragment.suraInput has correct value.

ozbek:
@ahmedre: we are in deadlock here, your input is highly appreciated :)

ahmedre:
sorry for the long delayed reply!
first off, jazakAllah khairan @fikr4n for bringing this up, this will be a great fix to have in sha' Allah!
i agree with @hakimrie here and tested his proposal and it seems to work (and is a much less change).

so i just added `layout.requestFocus();` in the block for 
```builder.setPositiveButton(getString(R.string.dialog_ok), (dialog, which) -> {```

and that seems to do the trick. what do you think? if it's good may you update your PR so we can merge this?

jazakAllah khairan.

fikr4n:
I am sorry, it was my mistake that I almost forget this PR.

> if it's good may you update your PR so we can merge this?

Just to confirm so I don't make another mistake, did you mean all of these?

1. Convert to Kotlin
2. Use `onForceComplete`, delete `forceComplete`
3. Add `focusableInTouchMode` and `clickable` in `jump_dialog.xml`

ahmedre:
@fikr4n jazakAllah khairan and no brother, please don't apologize, it's my fault for answering so late :(

i am suggesting something much simpler - if you completely start the patch over and just add `layout.requestFocus();` in the beginning of the code for when the person presses "ok" on the dialog, it should work.

the approach you described i think you tested and it didn't work for you.

jazakAllah khairan

fikr4n:
It's done, but not kotlinized yet.

fikr4n:
Is it okay now? Should we convert them to kotlin?

ozbek:
Looks good to me :+1:

ahmedre:
jazakAllah khairan and sorry it took so long!

fikr4n:
وجزاكم الله خيرا
Happy to contribute in what muslim bothers use everyday

PR #1027, opened 2018-12-15
Using AndroidX for Preference class

ahmedre:


ahmedre:
i just realized that i did a squash instead of a merge - i thought it would put "Ahmed committed with hakimrie" but apparently it didn't do so this time - didn't realize it otherwise would have done a merge instead - i am really sorry about that :(

ahmedre:
reverted this one and merged yours instead so you have credit now 😂 i'll merge my fixes now in sha' Allah

PR #999, opened 2018-10-30
Using AndroidX for Preference class

hakimrie:
I propose these changes because some of custom Preference no longer needed (eg. QuranPreference), now we can use sdk Preference class instead

Please review PR #995 first

PR #1028, opened 2018-12-15
Revert "Using AndroidX for Preference class (#1027)"

ahmedre:
This reverts commit a9ebf20f22b80343144ea70d67ac22146d855eb8.


ahmedre:
doing this to merge #999 instead and give @hakimrie credit - will apply my fixes separately.

PR #1057, opened 2019-01-29
Fix search's downloading of the Arabic database

ahmedre:


ahmedre:
this is a quick fix, but it's not the right fix, since this fix also was done on `PagerActivity` some time ago - the right fix is that the logic of whether to unzip or not should live in the `DownloadService` based on the downloaded url, but for now, this is a quick fix in preparation for a patch release.

PR #1098, opened 2019-03-15
Fix highlight overlapping

### LINKED ISSUE #951: highligh overlap
image_20_05_2018_925
image_20_05_2018_499
image_20_05_2018_189

in page 17

ahmedre:
This patch fixes highlight overlapping by removing rectangles that
overlap each other. It also changes the threshold of considering a line
to be a "full line" to be a percentage of the page size, which should
help fix some cases where the highlighting on the left or right side
does not form a straight vertical line. Fixes #951.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/data/AyahInfoDatabaseProvider.java
+++ b/app/src/main/java/com/quran/labs/androidquran/data/AyahInfoDatabaseProvider.java
@@ -1,12 +1,13 @@
 package com.quran.labs.androidquran.data;
 
 import android.content.Context;
-import androidx.annotation.Nullable;
 
 import com.quran.labs.androidquran.di.ActivityScope;
 import com.quran.labs.androidquran.util.QuranFileUtils;
 
 import javax.inject.Inject;
+
+import androidx.annotation.Nullable;
 
 @ActivityScope
 public class AyahInfoDatabaseProvider {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/quran/labs/androidquran/data/AyahInfoDatabaseProvider.java
```diff
@@ -31,4 +32,8 @@ public AyahInfoDatabaseHandler getAyahInfoHandler() {
     }
     return databaseHandler;
   }
+
+  public int getPageWidth() {
+    return Integer.valueOf(widthParameter.substring(1));
```
ahmedre: this is a bit of a hack, but it's the easiest way to do this at the moment. in the future, should just provide the actual width.

app/src/main/java/com/quran/labs/androidquran/model/quran/CoordinatesModel.java
```diff
@@ -25,7 +24,7 @@
 
 @ActivityScope
 public class CoordinatesModel {
-  private static final int PIXEL_THRESHOLD = 10;
+  private static final float THRESHOLD_PERCENTAGE = 0.015f;
```
ahmedre: 1.5% of the page width

app/src/main/java/com/quran/labs/androidquran/model/quran/CoordinatesModel.java
```diff
@@ -141,7 +174,7 @@ private AyahCoordinates normalizePageAyahs(AyahCoordinates ayahCoordinates) {
       if (lastRect.left < firstRect.right) {
         RectF middleBounds = new RectF(lastRect.left,
             /* top= */ firstRect.bottom,
-            firstRect.right,
+            Math.min(firstRect.right, lastRect.right),
```
ahmedre: instead of just taking `firstRect.right`, take the left most one. this fixes where the lines are drawn in very rare cases (like 1:7).

PR #1099, opened 2019-03-15
Start services as foreground sooner

ahmedre:
Whenever a service is started as foreground, make the startForeground
call (along with showing the notification) one of the first things that
the service does.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
+++ b/app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
@@ -357,6 +357,12 @@ public class AudioService extends Service implements OnCompletionListener,
     }
 
     final String action = intent.getAction();
+    if (ACTION_PLAYBACK.equals(action)) {
+      // this is the only action with startForegroundService, so
+      // go to the foreground as quickly as possible.
+      setUpAsForeground();
+    }
+
     if (ACTION_CONNECT.equals(action)) {
       if (State.Stopped == state) {
         processStopRequest(true);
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
```diff
@@ -357,6 +357,12 @@ public int onStartCommand(Intent intent, int flags, int startId) {
     }
 
     final String action = intent.getAction();
+    if (ACTION_PLAYBACK.equals(action)) {
```
ahmedre: just moving it up sooner - and start foreground even if we're already foreground (multiple calls are fine, but missing a call is asking for a crash).

app/src/main/java/com/quran/labs/androidquran/service/QuranDownloadService.java
```diff
@@ -233,6 +231,12 @@ private void sendNoOpMessage(int id) {
 
   @Override
   public int onStartCommand(Intent intent, int flags, int startId) {
+    // if it's a download, it wants to be a foreground service.
+    // quickly start as foreground before actually enqueueing the request.
+    if (ACTION_DOWNLOAD_URL.equals(intent.getAction())) {
```
ahmedre: this is actually a lot sooner than where it was happening before - before, `handleOnStartCommand` would process the message and then give it to the handler, which would eventually start foreground and show the notification. on some devices, this could take so long, so make it the first thing we do.

app/src/main/java/com/quran/labs/androidquran/service/util/QuranDownloadNotifier.java
```diff
@@ -273,6 +274,16 @@ public Intent notifyError(int errorCode, boolean isFatal, NotificationDetails de
     return progressIntent;
   }
 
+  public void notifyDownloadStarting(){
+    String title = appContext.getString(R.string.downloading_title);
+    notificationManager.cancel(DOWNLOADING_ERROR_NOTIFICATION);
+
+    lastMaximum = -1;
+    lastProgress = -1;
+    showNotification(title, appContext.getString(R.string.downloading_message),
```
ahmedre: most important here is the flag for foreground being explicitly set to true.

PR #1110, opened 2019-03-20
Minor fixes around file handling

ahmedre:


app/src/main/java/com/quran/labs/androidquran/QuranDataActivity.java
```diff
@@ -341,30 +342,6 @@ protected Boolean doInBackground(Void... params) {
...
-      // in 2.9.0, there was an initial release of the new madani images - since then, we
-      // moved to a new set of images that let us draw the ayah markers and sura headers
-      // (and are, consequently, much lighter). they also properly center in the pages,
-      // unlike the older ones. the number of people who get this update are very few, and
-      // so temporarily add some code to remove this directory for now. this code can be
-      // removed in a future version in sha' Allah. if this isn't removed, no harm is done,
-      // since the page size has changed anyway, and so only the new images will be used.
-      // the page sets are currently not compatible (since one doesn't have ayah markers
-      // and one does).
-      final File newMadani = new File(baseDir, "new_madani");
```
ahmedre: this used to delete the old new madani directory - it's not needed and it's been there for enough releases to hopefully have deleted those databases for people on the old version.

app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
```diff
@@ -150,10 +148,7 @@ public boolean haveAllImages(Context context, String widthParam, int totalPages)
         return true;
       } else {
         Timber.d("haveAllImages: couldn't find the directory, so making it instead");
-        makeQuranDirectory(context);
-        if (!IMAGES_DIRECTORY.isEmpty()) {
-          makeQuranImagesDirectory(context);
-        }
+        makeQuranDirectory(context, widthParam);
```
ahmedre: this is equivalent - with one exception that we're explicitly passing the width in instead of using the default width.

common/data/src/main/java/com/quran/data/pageinfo/common/size/DefaultPageSizeCalculator.kt
```diff
@@ -8,12 +8,12 @@ open class DefaultPageSizeCalculator(displaySize: DisplaySize) : PageSizeCalcula
   private var overrideParam: String? = null
 
   override fun getWidthParameter(): String {
-    return when {
+    return overrideParam ?: when {
```
ahmedre: this only matters for 1280 images (which is an override) where the 1024 would take precedence over the 1280, which is wrong in that case.

app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.java
ahmedre: should revisit this though because when coming from the code to check fallbacks, it'll end up creating at least 3 directories, most of which aren't needed

PR #1126, opened 2019-04-05
Update basmallah string

ahmedre:
This patch just updates the basmallah string used to match the one in
the uthmani v2 database.


ahmedre:
this is mostly to quick fix for a bug in which bookmarking the first ayah of any sura shows the basmallah on the bookmarks page. the better solution will be to standardize the database basmallah handling between v1 and v2 and use the Arabic table (instead of the sharing text table) for the bookmarks screen, but in sha' Allah will do this in the future.

Issue #1142, opened 2019-04-14
Navbar stays white when reading in night mode


MahmoodMahmood:
Bug confirmed in both oneplus 6 and samsung note 8.

Before:
![56970964_848470415488824_8782265241854541824_n](https://user-images.githubusercontent.com/14712160/56098093-dab4b400-5f0d-11e9-98a5-4405b9323112.jpg)

After:
![57133760_453967955347303_5333886062034944000_n](https://user-images.githubusercontent.com/14712160/56098208-2caa0980-5f0f-11e9-9e90-d53c99d1e3cf.jpg)

ahmedre:
should be fixed now, jazakAllah khairan for your patch.

PR #1143, opened 2019-04-14
Change navbar color in accordance with night mode (#1142)

MahmoodMahmood:


ahmedre:
jazakAllah khairan - i added some code to store the color when starting the activity and to restore it on light mode. since my phone's nav bar is always black, would appreciate if you'd give it a shot - i've gone ahead and merged this though, jazakAllah khairan!

PR #1156, opened 2019-04-27
Add many logs to track down disappearing images

ahmedre:


murtraja:
If I may ask, what is the q4a format? What data does it store?

ahmedre:
it's an empty file. when i detect that images are removed, i check if this file exists or not. if it doesn't exist, it means the entire directory was removed. if it does, it means only certain files were removed.

this is all to debug a complaint some people have that the app periodically deletes images and/or audio.

Issue #1148, opened 2019-04-21
Improve Arabic matching in fast jump screen


ahmedre:
Be nicer with completion in auto jump screen - for example, typing: الا does not match الأحقاف due to the hamza.

Ahmed9914:
Possible duplicate of #453 ?

MahmoudMabrok:
is it still open?

ahmedre:
this has already been fixed.

Issue #960, opened 2018-06-17
Arabic search engine is not working correctly 


3dola:
1) As an example, please look for the word tombs in Arabic (قبور) then you should get the error msg, "no results found for...قبور" 
2) Searching the Quran script in Arabic brings Quran results mixed with interpretation (tafseer) which makes the experience more difficult and less enjoyable

The search engine accuracy is a must have! 

Thank you for the time and efforts put into this app.

asyazwan:
For your first point it’s a known limitation:
#453

3dola:
This alone makes the app incompetent in its league. Primarily because it gives the average user a false impression about the results. 

In this case, maybe the search engine needs to be replaced/rewritten entirely.

ahmedre:
yes, absolutely - search is completely unacceptable today.

i intend on rewriting it in sha' Allah. part of it is our hands are slightly tied in the sense that things like stemming, etc don't work in the versions of SQLite that are on Android - but even for those, we have options of what we can do and improve both offline search (in addition to deciding to use online search when the connection is good).

Ahmed9914:
possible duplicate of #453

ahmedre:
closing because fixed.

Issue #953, opened 2018-06-01
"و" is not part of the arabic verb


ahadel:
In arabic "و" is speacial char like and in english 
You can see this issue by searching for this verb 
يقطعون 
The result will be 1 occurance and it should be 3 occurance ,

The missing 2 results is starting with "و" 
Wich is not part of the verb .

I think the seaech algorithm should not search for the whole word , part of the word is correct , i see the same problem with starting "ال"  for each word in search term , "ال"  example "الجنة" and "جنة"

جزاكم الله خيرا 
أحمد

ahmedre:
closing because fixed.

Issue #453, opened 2014-10-01
Improving search


asyazwan:
In reference to #449 and #447.

The search for Quran text currently works in mysterious way (ie. unclear). For example searching for `لل` doesn't include `الله` but will include `لله`.

I would like to improve this to be more intuitive as part of my academic project.

Areas to check:
- Search with regex (is this too much?)
- Search with/without diacritic
- Search scope option (like which translations)

I have not delved into the search code so I can't comment on the technical feasibility yet. Any advice is appreciated, especially on the UI aspect.

ahmedre:
these solutions are too much work to be worth doing imo due to the complexity of these solutions - i think that the best idea is to use the new quran.com search api when the device is connected.

@mmahalwy do you think alpha's quran search api is stable enough to include in quran android now or in the near future?

ahmedre:
see also #427 about some examples of poor search quality.

mmahalwy:
@ahmedre yeah its stable enough now. I think the improvement is the transliteration which I am waiting from you :)

ahmedre:
we should also search sura names.

ahmedre:
from email:
search for any word in basmallah returns the first ayah of most suras

m7mdyahia:
> another solution (just for documentation purposes) is bundling our own sqlite with the app with the unicode stemmer enabled (if it works, otherwise we may need our own Arabic stemmer) 

Unicode stemmer will not work in arabic as you expect it will not remove diacritic also it will not make equivlance between characters as (ي) and (ى) this is because unicode definittion for arabic language don't consider removing diacritic from text or matching similar letters as [normalization or canonical equivalence](https://en.wikipedia.org/wiki/Unicode_equivalence)

this could be tested using a newer version of sqlite on pc (I would appreciate if someone confirmed this)

so I think generally (for any arabic language search using fts sqlite)  for arabic language we need to write sqlite arabic tokinizer then see what could be done to ship it for android (I know this is alot of work but it will be very helpfull to many applications)

sneetsher:
Also  #268 quiet related. I agree with @ahmedre searching Arabic is vast topic and need to many underlying tools & libraries.

The best is to relay on another ready project that provide such feature for off-line search or if there no perfect solution some on-line API. Like the one mentioned quran.com.

I would suggest that you make a generic interface that allows you to switch search provider later for an off-line solution. 

[...]

PR #1164, opened 2019-04-28
Fix Arabic search

### LINKED ISSUE #427: poor search quality
from an email:
- When search "idhrib" (اضرب) it doesn't show results about "wa
  idhrib" (واضرب), and when search "dharaba" (ضرب), it doesn't show
  results about "madhrib" (مضرب)
- When search "wa" (و) it doesn't show more than 150 results although
  there is "22,226" result for this phrase.

ahmedre:
Al7amdulillah, this patch completely rewrites (and fixes) Arabic search.
It now supports searching with approximate matches for characters that
could have multiple variants in Arabic (i.e. an alif now matches an alif
with a hamza, etc), among other things. It doesn't support tashkeel
search yet, but in practice, this isn't a problem since Arabic search is
only used for searching the Quran text without taskeel.

Fixes #427, #453, #953, and #960.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -107,6 +107,7 @@ dependencies {
   implementation project(path: ':common:data')
   implementation project(path: ':common:networking')
   implementation project(path: ':common:pages')
+  implementation project(path: ':common:search')
 
   implementation deps.kotlin.stdlib
   implementation "androidx.appcompat:appcompat:${androidxAppcompatVersion}"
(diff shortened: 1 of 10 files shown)


ahmedre:
special thanks to @mustafa0x for the replacement characters here.

Ahmed9914:
Fixes #1148

ahmedre:
not yet actually - this patch is only for search itself, the other ticket is for jump

Ahmed9914:
@ahmedre 
You mean another regex matching expression should be added in this block:
https://github.com/quran/quran_android/blob/636396fd9350f3dfe0b8f4e474da3aa06979c3a4/app/src/main/java/com/quran/labs/androidquran/ui/fragment/JumpFragment.java#L102

Is this right?

ahmedre:
yes

PR #1189, opened 2019-05-26
Feature delete media

murtraja:
Phase 1: gapless
@ahmedre, I couldn't find any non-gapless qaris so I couldn't implement the non-gapless download and delete. How can I enable non-gapless qaris?

I have tested this with Husary. The PR allows download/delete of individual surah of a particular qari. I would love to have a thorough review before I proceed for UI changes (which you can point out too)

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
@@ -91,6 +91,18 @@ public class AudioManagerActivity extends QuranActionBarActivity
     if (disposable != null) {
       disposable.dispose();
     }
+    mOnDownloadInfo = new DisposableSingleObserver<List<QariDownloadInfo>>() {
+      @Override
+      public void onSuccess(List<QariDownloadInfo> downloadInfo) {
+        progressBar.setVisibility(View.GONE);
+        shuyookhAdapter.setDownloadInfo(downloadInfo);
+        shuyookhAdapter.notifyDataSetChanged();
+      }
+
+      @Override
+      public void onError(Throwable e) {
+      }
+    };
     disposable = AudioManagerUtils.shuyookhDownloadObservable(quranInfo, basePath, qariItems)
         .observeOn(AndroidSchedulers.mainThread())
         .subscribeWith(mOnDownloadInfo);
@@ -120,19 +132,7 @@ public class AudioManagerActivity extends QuranActionBarActivity
     super.onDestroy();
   }
 
-  private DisposableSingleObserver<List<QariDownloadInfo>> mOnDownloadInfo =
-      new DisposableSingleObserver<List<QariDownloadInfo>>() {
-        @Override
-        public void onSuccess(List<QariDownloadInfo> downloadInfo) {
-          progressBar.setVisibility(View.GONE);
-          shuyookhAdapter.setDownloadInfo(downloadInfo);
-          shuyookhAdapter.notifyDataSetChanged();
-        }
-
-        @Override
-        public void onError(Throwable e) {
-        }
-      };
+  private DisposableSingleObserver<List<QariDownloadInfo>> mOnDownloadInfo;
 
   private View.OnClickListener mOnClickListener = new View.OnClickListener() {
     @Override
--- a/app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
@@ -139,12 +139,8 @@ public class AudioManagerActivity extends QuranActionBarActivity
     public void onClick(View v) {
       int position = recyclerView.getChildAdapterPosition(v);
       if (position != RecyclerView.NO_POSITION) {
-//        QariDownloadInfo info = shuyookhAdapter.getSheikhInfoForPosition(position);
-//        if (info.downloadedSuras.size() != 114) {
-//          download(qariItems.get(position));
-//        }
         Intent intent = new Intent(AudioManagerActivity.this, SurahAudioManager.class);
-        intent.putExtra("Sheikh", position);
+        intent.putExtra(SurahAudioManager.EXTRA_SHEIKH_POSITION, position);
         startActivity(intent);
       }
     }
--- a/app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
@@ -1,15 +1,14 @@
 package com.quran.labs.androidquran.ui;
 
-import android.content.Intent;
-import android.content.IntentFilter;
-import android.os.Bundle;
-import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 import androidx.appcompat.app.ActionBar;
+import androidx.appcompat.app.AppCompatActivity;
 import androidx.recyclerview.widget.DefaultItemAnimator;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
+
+import android.content.Intent;
+import android.os.Bundle;
 import android.view.LayoutInflater;
-import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
@@ -20,11 +19,7 @@ import com.quran.labs.androidquran.QuranApplication;
 import com.quran.labs.androidquran.R;
 import com.quran.labs.androidquran.common.QariItem;
 import com.quran.labs.androidquran.data.QuranInfo;
-import com.quran.labs.androidquran.data.SuraAyah;
-import com.quran.labs.androidquran.service.QuranDownloadService;
 import com.quran.labs.androidquran.service.util.DefaultDownloadReceiver;
-import com.quran.labs.androidquran.service.util.QuranDownloadNotifier;
-import com.quran.labs.androidquran.service.util.ServiceIntentHelper;
 import com.quran.labs.androidquran.util.AudioManagerUtils;
 import com.quran.labs.androidquran.util.AudioUtils;
 import com.quran.labs.androidquran.util.QariDownloadInfo;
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
```diff
@@ -139,10 +139,13 @@ public void onError(Throwable e) {
     public void onClick(View v) {
       int position = recyclerView.getChildAdapterPosition(v);
       if (position != RecyclerView.NO_POSITION) {
-        QariDownloadInfo info = shuyookhAdapter.getSheikhInfoForPosition(position);
-        if (info.downloadedSuras.size() != 114) {
-          download(qariItems.get(position));
-        }
+//        QariDownloadInfo info = shuyookhAdapter.getSheikhInfoForPosition(position);
```
ahmedre: please just remove instead

app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
```diff
@@ -139,10 +139,13 @@ public void onError(Throwable e) {
...
+//        if (info.downloadedSuras.size() != 114) {
+//          download(qariItems.get(position));
+//        }
+        Intent intent = new Intent(AudioManagerActivity.this, SurahAudioManager.class);
+        intent.putExtra("Sheikh", position);
```
ahmedre: can we add a constant for this somewhere?

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+    final ActionBar ab = getSupportActionBar();
+    if (ab != null) {
+      ab.setTitle(R.string.audio_manager);
+      ab.setDisplayHomeAsUpEnabled(true);
+    }
+
+    setContentView(R.layout.activity_surah_audio_manager);
+
+    Intent intent = getIntent();
+    sheikhPosition = intent.getIntExtra("Sheikh", -1);
```
ahmedre: when you replace that with a constant you can just refer to the constant here.

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+      finish();
+      return true;
+    }
+    return super.onOptionsItemSelected(item);
+  }
+
+  private void getShuyookhData() {
+    if (disposable != null) {
+      disposable.dispose();
+      mOnDownloadInfo =
```
ahmedre: can we rename this variable to not have an `m` at the start?

ahmedre: actually why do we need this and the other one that's set on line 126? can we just inline one of them directly in the subscribe instead?

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+        boolean downloaded = info.downloadedSuras.get(surah);
+        if(downloaded) {
+          // TODO: show a confirmation dialog before deleting
+          // delete it
+          delete(surah);
+        } else {
+          // download
+          download(surah);
+        }
+//        if (info.downloadedSuras.size() != 114) {
```
ahmedre: remove if we don't need it

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+
+  private void delete(int surah) {
+    QariItem qariItem = qariItems.get(sheikhPosition);
+    String baseUri = basePath + qariItem.getPath();
+    String fileUri = audioUtils.getLocalQariUri(this, qariItem);
+    String fileName = fileUri.format(Locale.US, fileUri, surah);
+    File audioFile = new File(fileName);
+    String resultString;
+    if(audioFile.delete()) {
+      resultString = "Successfully deleted the file";
```
ahmedre: use strings.xml

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+    getShuyookhData();
+  }
+
+  @Override
+  public void handleDownloadFailure(int errId) {
+    getShuyookhData();
+  }
+
+  private class SurahAdapter extends RecyclerView.Adapter<SurahAudioManager.SurahViewHolder> {
+    private final LayoutInflater mInflater;
```
ahmedre: please don't use `m` at the start of variable names

app/src/main/java/com/quran/labs/androidquran/ui/SurahAudioManager.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    @Override
+    public SurahAudioManager.SurahViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+      return new SurahAudioManager.SurahViewHolder(mInflater.inflate(R.layout.audio_manager_row, parent, false));
+    }
+
+    @Override
+    public void onBindViewHolder(SurahAudioManager.SurahViewHolder holder, int position) {
+      holder.name.setText((position+1) + "");
```
ahmedre: should we show the sura name here instead of the number?

murtraja:
> 
> 
> jazakAllah khairan for doing this and sorry for the late reply - might be a bit late replying these days, i apologize in advance.

only a couple days left now :)

ahmedre:
jazakAllah khairan - this looks good. should we wait until we support gapped and "download all" before merging? what do you think @murtraja ?

murtraja:
@ahmedre, of course. Can you help how to enable gapped qaris in Madani Debug?
P.S. I also need inputs on the UI, if you can provide some pointers, I can also change the UI

ahmedre:
i just merged a PR that adds a gapped qari so you can try now.
for the ui, the bookmarks ui is around what i was thinking of - so basically:

1. there'd be a button in the toolbar to download all
2. long pressing any item would "select it" and change the toolbar to have 2 buttons - a deletion x and a download button.

try adding a bunch of bookmarks and long pressing one of them in the index screen and you'll see what i mean - let me know what you think though. jazakAllah khairan.

murtraja:
I agree. for the 2nd point, I will additionally only enable download button if atleast 1 of the selections is not downloaded. Similarly I will activate the delete button when there is at least 1 selection which is already downloaded and can be deleted.

This brings us to a complex case of gapped qari where it is also possible to have a partially downloaded surah. in this case if user long presses on a partially downloaded gapped surah, both delete and download will be available

Also, I will have to pick good icons to denote the said features.

ahmedre:
for gapped, let's make download and delete operations act on the entire sura (i.e. let's not split it up by one entry per ayah). this will make it simpler.

for icons, use something from Google - https://material.io/tools/icons/?style=baseline

murtraja:
> i just merged a PR that adds a gapped qari so you can try now.
I have rebased this branch off of master after your gapped commit, I hope this is the expected work flow. I didn't want to introduce a new merge commit :) Is it fine?

murtraja:
Work done so far:
1. Adding another activity that allows download/delete of individual Surah of Sheikhs (instead of downloading all Surah)

Todo:
1. download/delete all option
2. long click selection of surah for bulk download / delete
3. UI changes - icons, changing header from normal to selection mode
4. Make download non blocking

murtraja:
s1
s2

murtraja:
s3
s4
s5

murtraja:
Work done so far:
1. Individual Surah download status for each Qari
2. Individual download/delete Surah
3. Selection download/delete - Only UI

Todo:
1. Selection download/delete - functionality

ahmedre:
salam 3alaikum,
just tried this again today - Masha’Allah, it is looking good. one minor issue i found - if you make changes in a particular sheikh's screen (i.e. remove an audio file or add one), the main list (showing all the shuyookh) doesn't refresh.

will wait until you're done and then review the code in sha' Allah.
jazakAllah khairan for doing this!
walsalam 3alaikum.

app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
```diff
@@ -91,6 +91,18 @@ private void getShuyookhData() {
     if (disposable != null) {
       disposable.dispose();
     }
+    mOnDownloadInfo = new DisposableSingleObserver<List<QariDownloadInfo>>() {
```
ahmedre: let's use a local variable here instead

app/src/main/java/com/quran/labs/androidquran/ui/AudioManagerActivity.java
```diff
@@ -120,29 +132,16 @@ protected void onDestroy() {
...
-          progressBar.setVisibility(View.GONE);
-          shuyookhAdapter.setDownloadInfo(downloadInfo);
-          shuyookhAdapter.notifyDataSetChanged();
-        }
-
-        @Override
-        public void onError(Throwable e) {
-        }
-      };
+  private DisposableSingleObserver<List<QariDownloadInfo>> mOnDownloadInfo;
```
ahmedre: and let's remove this

app/src/main/java/com/quran/labs/androidquran/util/AudioUtils.kt
```diff
@@ -75,6 +75,16 @@ constructor(private val quranInfo: QuranInfo, private val quranFileUtils: QuranF
     return if (rootDirectory == null) null else rootDirectory + item.path
   }
 
+  fun getLocalQariUri(context: Context, item: QariItem): String? {
```
ahmedre: why not just use `getQariUrl` in this same class?

murtraja: I actually used that initially but found that the method uses `"%03d%03d"` but instead it should be `"%03d" + File.separator + "%03d"`. So I wasn't sure if I could change the existing `getQariUrl` method without breaking existing code so I skipped it and wrote a new function altogether. Would you like me to use `getQariUrl` and make the necessary modifications so that it works correctly for gapless?

murtraja: ohh actually, `getQariUrl` gives me the remote URL, I want a local URL so that I can find the file and delete. So please ignore my above comment

murtraja:
> 
> 
> salam 3alaikum,
> just tried this again today - Masha’Allah, it is looking good. one minor issue i found - if you make changes in a particular sheikh's screen (i.e. remove an audio file or add one), the main list (showing all the shuyookh) doesn't refresh.
> 
> will wait until you're done and then review the code in sha' Allah.
> jazakAllah khairan for doing this!
> walsalam 3alaikum.

Well, I did observe this problem and thought of using BehaviorSubject to overcome the issue. It will allow multiple subscribers (AudioManager and SurahAudioManager) and whenever we subscribe, it will give the last recent data (http://reactivex.io/documentation/subject.html)

I am not sure how disposing will work in this case, but I am reading more about it and will have more visibility after my initial study is completed.

murtraja:
Work done so far (If I remember correctly):
1. individual + multiple download/delete done

Todo:
1. BUG - Updating the main qari selection screen, when the child surah selection gets updated
2. Make downloads non-blocking

ahmedre:
masha'Allah, great! looking forward to the rest of your changes.

murtraja:
Should I go ahead with the Subject approach which I have mentioned previously in order to fix the bug? What are your thoughts?

ahmedre:
that could work - but might make sense to cache the observable instead and add `.share()` to the end so that it's reusable in multiple places.

ahmedre:
jazakAllah khairan - i went ahead and fixed some of the issues with this and cleaned it up a bit and merged it. didn't implement background download yet, but it's not blocking for releasing this. really appreciate your work on this @murtraja

murtraja:
I am glad I could be of help. I am sorry for not completing the task :(

PR #1213, opened 2019-10-18
fix weird behaviour in cutout mode

sodaoud:


ozbek:
Can you describe the "weird behaviour" you are experiencing without this change? I see no anomalies on my device with a cut-out at the top.

sodaoud:
the default cutout mode lets the activity display being cut by the notch, in the case of the PagerActivity, when the user clicks the screen, it switches the toolbar from visible to hidden, when in transition the activity display changes going all the way up, to just under the notch. which makes it look "jumping" a little in the transition.

The fix I propose, is to prevent the from happening, and letting the PagerActivity displays "always" under the notch.

you can see the difference in display here: https://developer.android.com/guide/topics/display-cutout#best_practices_for_display_cutout_support

ozbek:
I see, you meant the jumping/jittering of the screen up and down. Good catch!

Cc: @ahmedre

ahmedre:
jazakAllah khairan for this!

Issue #985, opened 2018-10-04
Making a custom entries in the number of ayah looping


fnzainal:
I love to use this application to listen to the Quran and its very helping for memorizing quran, Alhamdulillah. And the number of ayat repetitions is very limited. The current version, there is some options of repetitions (1,2,3 and loop)

And I want to make a custom entries in the number of ayat repetitions. 
I imagine, the design of the form will be like "Jump" feature, when we fill in the surah name / number of ayat.

I think it will help someone who trying to memorize the quran.

PR #1220, opened 2019-10-24
Improve minor feature for repeat range for issue #985

fnzainal:
Just make a little change for my issue #985 .
Sometimes for helping memorize Quran need more repeat range, so I add 4,5,10,15,20 for this issue.
Hope it helpful for me and others. 

Thanks. Baarakallahufiikum

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
@@ -24,7 +24,7 @@ import com.quran.labs.androidquran.widgets.QuranSpinner;
 import javax.inject.Inject;
 
 public class AyahPlaybackFragment extends AyahActionFragment {
-  private static final int REPEAT_MAX = 3;
+  private static final int REPEAT_MAX = 8;
   private static final int ITEM_LAYOUT = R.layout.sherlock_spinner_item;
   private static final int ITEM_DROPDOWN_LAYOUT = R.layout.sherlock_spinner_dropdown_item;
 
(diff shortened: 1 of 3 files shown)


ozbek:
This is essentially same as the existing user experience (albeit 8 options instead of 4), still limited options.

ahmedre:
I agree with @ozbek - this isn’t great for ui and still doesn’t fully solve the problem. Let’s try to think of a better solution - maybe an “other” option that lets the person choose how many times?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

app/src/main/res/values/strings.xml
```diff
@@ -336,6 +336,11 @@
     <item>1 time</item>
     <item>2 times</item>
     <item>3 times</item>
+    <item>4 times</item>
+    <item>5 times</item>
+    <item>10 times</item>
+    <item>15 times</item>
+    <item>20 times</item>
```
ahmedre, line 343: can you add these to the other translation files?

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
```diff
@@ -245,16 +245,29 @@ private void updateEnforceBounds(int rangeRepeatPosition) {
...
-      return repeat;
+    switch (repeat) {
+      case -1:
+        return REPEAT_MAX;
+      case 9:
+        return 5;
+      case 14:
+        return 6;
+      case 19:
+        return 7;
```
ahmedre, line 256: why?

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
```diff
@@ -245,16 +245,29 @@ private void updateEnforceBounds(int rangeRepeatPosition) {
...
 
   private int positionToRepeat(int position) {
     if (position >= REPEAT_MAX) {
       return -1;
+    } else if (position == 5){
+      return 9;
+    } else if (position == 6){
+      return 14;
+    } else if (position == 7){
+      return 19;
```
ahmedre, line 270: do we need this?

ahmedre:
though i'd love a better solution, i am ok taking this if you can help clean it up - jazakumAllah khairan for your patience and for your help with this.

ahmedre:
also can we just make the new options 5 and 10?

benomaire:
Can we use a [horizontal number picker](https://github.com/ShawnLin013/NumberPicker) in the settings? and only add 5 and 10 for example for the lower icon as @ahmedre suggested?

benomaire:
@fnzainal Do you allow me to carry on this work?

ahmedre:
closing now since @benomaire finished it in #1519

ahmedre:
jazakumAllah khairan!

Issue #1210, opened 2019-09-23
Connection issue


Bakhrom:
When using in Uzbekistan, there are problems with the download, it probably appeared after blocking Facebook in 2018. When using custom dns (1dot1dot1dot1.cloudflare-dns.com), the application works with mobile internet. Otherwise, only with VPN. Is there any way to solve this problem from the application side?

ibragimof:
The same problem. Please, fix it for users from Uzbekistan.
![Screenshot_2019-12-02-17-18-58-461_com quran labs androidquran](https://user-images.githubusercontent.com/47029060/69982267-d9e09900-1555-11ea-9ff6-0a5fc3351b5f.jpg)
![Screenshot_2019-12-02-17-17-08-995_com quran labs androidquran](https://user-images.githubusercontent.com/47029060/69982268-da792f80-1555-11ea-8b53-4eb77fa6ef67.jpg)

ozbek:
ATM, [setting up Cloudflare DNS](https://developers.cloudflare.com/1.1.1.1/setting-up-1.1.1.1/android/) by end users seems to resolve the issue.

ibragimof:
Updating in client side not resolved the problem. Still connection possible via vpn (and only).

ahmedre:
salam 3alaikum,
@ibragimof i just set up a test that goes to cloudflare after the first failure so that the files download through cloudflare instead of directly through our servers - may you try this apk out (it should be installable side by side with the production one so you don't lose any data) and see if you can download the initial set of images with it and let me know?

http://cafesalam.net/uploads/quran-test-cloudflare.apk

if it doesn't work, may you go to settings -> advanced -> send logs and send us the logs?
jazakumAllah khairan,
walsalam 3alaikum

ozbek:
ping @Bakhrom

Bakhrom:
Page loading working, but audio, translations, "Page Type (experimental)" not working
[Logs.txt](https://github.com/quran/quran_android/files/5722574/Logs.txt)

![scr1](https://user-images.githubusercontent.com/9442259/102744764-5ff79980-437c-11eb-9ee0-51c9630a0365.jpg)
![scr2](https://user-images.githubusercontent.com/9442259/102744770-6128c680-437c-11eb-9568-e9408d4f25e1.jpg)
![src3](https://user-images.githubusercontent.com/9442259/102744775-62f28a00-437c-11eb-89a7-979fd95b36ed.jpg)
[...]

PR #1233, opened 2019-12-15
Update dns fallback and add Cloudflare

ahmedre:


ozbek:
Would this fix #1210?

ahmedre:
sadly no - it was my first try for them though but the brother tried it and said it didn't help (seems like they block both based on url and via dns whereas before it was dns only). i am trying to add a fallback to another domain that i'll send to him to see if it fixes the problem (at least until they block it again :().

PR #1240, opened 2020-01-09
Fix toolbar being cut on api 28 and above

ahmedre:


app/src/main/res/values-v28/themes.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <style name="QuranToolBar.Overlay">
+        <item name="windowActionBarOverlay">true</item>
```
ahmedre: this file was added in #1213 but without copying the existing flags from `values-v19`. also, the file should be named themes.xml for consistency.

PR #1249, opened 2020-01-17
Remove Google's Dns Over Https for now

ahmedre:


ahmedre:
The url changed to dns.google/dns-query, yet in testing on the emulator, I noticed that I'd often get a 404 during lookup (despite the URL built by OkHttp being correct when run on curl, for example). Removing it for the time being.

PR #1264, opened 2020-02-01
Update proguard rules for dnsjava

ahmedre:


app/proguard.cfg
```diff
@@ -2,7 +2,6 @@
 -renamesourcefileattribute SourceFile
 -keepattributes *Annotation*
 -keepattributes SourceFile,LineNumberTable
--repackageclasses ''
```
ahmedre: added this long ago to work around a support library bug that's probably fixed now.

Issue #1252, opened 2020-01-19
Re-add partial page clean up text


ahmedre:
There used to be partial page clean up code to clean up cases where people downloaded the pages when we had bugs (resulting in a potentially partial page). Still getting reports about partial pages so really should add this code back.

ahmedre:
Also trying to consider bundling the pages with the app. This would make it easier to “fix” partial pages by overriding them and also would “fix” some of the network issues and download errors people constantly complain about. It can also help with scoped storage.

One point is what pages would make sense to bundle - new or old madani? Right now leaning more towards old madani to help migrate people with issues and partial pages now, and in a future update, make new madani the default and potentially show the page type switcher on first launch.

ozbek:
> consider bundling the pages with the app

Why wasn't that done in the beginning? What has changed since?

ahmedre:
it wasn't done early on since we had several different densities that we offered devices and tried to give each device the "closest" density for their screen. ever since, we've moved to a world in which most people with old madani are getting the same size of images for a while now, and all people on new madani are getting images of the same width.

ahmedre:
also, combining this with the complaints about partial pages makes this more interesting of an idea.

for history of the partial page issues, see #991 with the fix for partial page downloads (which i am hoping is effective and there's no other edge case for this), and #1018 (which used to delete on startup, but removed it after a few months thinking most people likely migrated).

ahmedre:
the discussion of bundling images is still one i am thinking about, but probably not for this next release that i aim to push out in the upcoming days in sha' Allah.

benomaire:
A recent comment on play store (11 Sep 2020)

> تطبيق ممتاز جدا جدا بس فيه الصفحه الثالثه في سوره مريم نصفها مفقود

Do you know what exactly the problem is now? I mean, is it just incomplete download?
Can we require hashes.txt (md5) next to images and compare them at every app start?
or may be compare image sizes with known values (should be faster)?

PR #1267, opened 2020-02-02
Re-remove the partial page checker

ahmedre:
The partial page checker was added back due to some remaining reports
about partial pages existing (refs #1252). In trying this on an actual
device, this takes a good several seconds on the first load (though in a
background thread), which isn't a great experience.


ahmedre:
Initially, I tried to optimize the checker (changed the method to only read a quarter of the pixels from each of the selected rows instead of the full row) - that actually didn't make too big a difference in performance from the quick tests I ran (most likely because the image, scaled down to 1/16th, is already really small, and the cost of `getPixels` is probably the vast amount of that, and it stops at the first pixel it finds among the last several rows, which is likely very soon in the checking process).

Looking back at the numbers, it seems that a good number of people had actually gotten the releases that ran images through this checker and its code. At this point, that leaves two options - let everyone suffer the ~5 second hit on startup to help the few people who never got that upgrade (perhaps showing a progress dialog to make their experience better), or remove the check and help people who complain to download an older version. Seeing as though there are relatively few complaints, I think it's a safe option to not make the vast majority of people have to pay this cost.

One potential idea is when starting to think about supporting scoped storage, bundle the legacy images and always copy from the bundled images to internal storage (as opposed to copying from the sdcard to scoped storage), which should fix the problem for the few remaining people (at least until the images are no longer bundled :)).

ahmedre:
i am still debating over this - will sleep on it in sha' Allah and see.

@ozbek interested to know your opinion on this - drop the checker or add a "loading" dialog and keep it for another few releases.

ahmedre:
was talking to @mustafa0x who suggested trying crc32/md5 to see if they help - will give it a shot and see how quick they are in comparison to this approach.

ahmedre:
closing this for now since am thinking about alternative solutions.

PR #1281, opened 2020-02-21
Support playing an ayah from translation

ahmedre:


nacer80:
جزاك الله خيرا أخي الكريم

PR #1283, opened 2020-02-23
Force re-init of overlayParams on view invalidate

ismail-s:
I noticed a bug where, when toggling night mode, the overlay text at the top of Quran Arabic pages does not change colour for currently active fragments (ie the fragment currently displayed and one either side). In particular, when going from night mode to normal mode, this currently makes the text unreadable.

I'm not sure if this is the best way to fix this, as, from debugging this, there are several calls to `invalidate` where this code gets run but doesn't need to. That said, this felt cleaner to me than adding the equivalent code to the end of this method:

https://github.com/quran/quran_android/blob/44715eaaf2761b5560154f88affe1f57a7bab2f8/app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java#L157-L173

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
+++ b/app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
@@ -266,6 +266,14 @@ public class HighlightingImageView extends AppCompatImageView {
     didDraw = true;
   }
 
+  @Override
+  public void invalidate() {
+    super.invalidate();
+    if (overlayParams != null) {
+      overlayParams.init = false;
+    }
+  }
+
   private Paint getPaintForHighlightType(HighlightType type) {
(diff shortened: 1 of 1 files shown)


ismail-s:
Tests pass locally for me. I don't think I can rerun the CI build.

ahmedre:
jazakAllah khairan - this is a good bug!

suggestion - what if instead we fix it in `setNightMode` itself, which is called - so right now, there's a block in `initOverlayParams` that looks like:

```java
    int overlayColor = overlayTextColor;
    if (isNightMode) {
      overlayColor = Color.rgb(nightModeTextBrightness,
          nightModeTextBrightness, nightModeTextBrightness);
    }
    overlayParams.paint.setColor(overlayColor);
```

can we extract this in a method and just call it from `setNightMode`? i think that'd fix the problem and would be cleaner.

PR #1287, opened 2020-02-26
Long press play should play the ayah directly

ahmedre:
Instead of showing the dialog when someone long presses an ayah and
chooses play, start playing right away. The dialog can be reached from
the settings button on the bottom bar when needed.


nacer80:
Jazek allaho khiren 
Good idea 
We hope changing of color of aya tefsir from dark to moshaf color (like yellow color with black font).
and aslo change of name of tefsir to arabic if the language of app arabic because always appears with English for example  sa3di tafseer  تفسير السعدي

PR #1290, opened 2020-03-14
Update gradle version and dependencies

ahmedre:
Also remove some unnecessary utility scripts.


build.gradle
```diff
@@ -39,7 +41,7 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:3.6.0'
+    classpath 'com.android.tools.build:gradle:4.0.0-beta02'
```
ozbek: Is there any reason to stay on bleeding edge? :)

I remember there was a need to use beta version because stable one did not support module-ing at that time. If that's not a case, I'd recommend staying with the stable releases.

ahmedre: there's a 4.1 canary 2 out so this isn't *very* bleeding edge 😂

PR #1295, opened 2020-03-24
Fix missing translation headers

### LINKED ISSUE #1291: Translation Names sometimes don't show
Got more than one a bug reports with screenshots - the dropdown has the items checked and the names are there, but in both the inline and fullscreen translation views, the headers for the translation type aren't showing.

ahmedre:
Fixes #1291.


nacer80:
While the application in arabic mode but tafsir and translation head still in English. 
Ex: Sa3di tafseer not changes to تفسير السعدي

ahmedre:
jazakAllah khair - fixed in #1310 - though you'll have to remove the tafseer and re-add it for the name to show up properly after that patch.

Issue #1201, opened 2019-08-08
Missing space in Arabic text in translation mode [al-Hijr:7]


mbarbar:
As-salam alaykum,

In surat al-Hijr (using the app's copying functionality)...

Translation mode:
لَّوۡمَا تَأۡتِينَا بِٱلۡمَلَٰٓئِكَةِ إِن كُنتَ مِنَ ٱلصَّٰدِقِينَ

(Sahih International)
Why do you not bring us the angels, if you should be among the truthful?"

-Sura Al-Hijr, Ayah 7

Mushaf mode:
(لَّوۡ مَا تَأۡتِینَا بِٱلۡمَلَـٰۤىِٕكَةِ إِن كُنتَ مِنَ ٱلصَّـٰدِقِینَ)
[Surat Al-Hijr 7]

As you can see, a space is missing after the first word.

mbarbar:
I also noticed after posting, copying in one mode gives you "Sura", and in another you get "Surat".

ahmedre:
wa3laikum alsalam,
jazakAllah khairan for the reports. will fix these in sha’ Allah.
walsalam 3alaikum.

mustafa0x:
لوما is sometimes spelled with a space and sometimes without (even by QPC). It makes little difference either way. This isn't worth changing now. If it later becomes clear that one of the two approaches has greater "support", then this change can be made.

ahmedre:
closing for now.

PR #1301, opened 2020-03-27
Normalize the usage of sura/surat/surah as surah

ahmedre:
Refs #1201.


murtraja:
The database too has "sura_number", have you normalized that as well?

PR #1313, opened 2020-04-10
Floating highlights

murtraja:
Pros:
1. Ability to animate highlights from one ayah to another.
2. If current playing ayah changes while the animation is in progress, then the current "in-between" highlight is morphed to the changed ayah highlight
3. If user has selected ayahs in ayahMode then animation is not shown

Cons:
1. Very inefficient code
2. Animation not scalable to other highlight types should we decide to add animation later to types other than AUDIO. May require quite a lot of refactoring later.
3. Needs thorough review
4. Lots of TODOs


murtraja:
@ahmedre Did you get a chance to review this? Let me know the problems in comments so that I can fix the issues. (I have mentioned some issues in comments/Todos in the code itself, but haven't fixed it myself because I need your inputs on those)

ahmedre:
really sorry akhi :( i will look at this soon in sha' Allah - jazakAllah khairan for your efforts on this and on all the other things you've done and please forgive me

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
   }
 
-  public void highlightAyah(int sura, int ayah, HighlightType type) {
+  private void highlightFloatableAyah(Set<String> highlights, String currentSurahAyah) {
+    final Map<String, List<AyahBounds>> coordinatesData = ayahCoordinates.getAyahCoordinates();
+
+    String previousSurahAyah = currentSurahAyah;
+    for(String surahAyahs: highlights) {
+      previousSurahAyah = surahAyahs;
+    }
```
ahmedre, line 161: this loop will basically just get the "last" one, and that perhaps can change - what are you actually trying to get here then?

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+    if(arrowIndex != -1) {
+      // The ayah changed during animating
+      startingBounds = (List<AyahBounds>)animator.getAnimatedValue();
+      animator.cancel();
+      previousSurahAyah = previousSurahAyah.substring(arrowIndex + 2);
+    } else {
+      startingBounds = coordinatesData.get(previousSurahAyah);
+    }
+
+    final String ayahTransition = previousSurahAyah+"->"+currentSurahAyah;
```
ahmedre, line 173: i'd vote for an object - maybe `AyahTransition` or something that has a from and to instead of manually having an arrow and parsing it like this, since it's more complicated that way

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+
+    // yes we make copies, because normalizing the bounds will change them
+    List<AyahBounds> previousAyahBoundsList = new ArrayList<>(startingBounds);
+    List<AyahBounds> currentAyahBoundsList = new ArrayList<>(coordinatesData.get(currentSurahAyah));
+
+    highlights.clear();
+    highlights.add(ayahTransition);
+
+    // TODO: reason whether this TypeEvaluator should be anonymous or not
```
ahmedre, line 182: i'd vote for not because:
1. this code doesn't ideally belong in HighlightingImageView
2. you can write a nice unit test for it that way without needing a ui test just to test the behavior (i.e if your standalone class exposes these as functions, and the animation stuff is outside that class)

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+        int minSize = Math.min(startSize, endSize);
+        int maxSize = Math.max(startSize, endSize);
+        List<AyahBounds> minList = startSize < endSize? start : end;
+        int diff = maxSize - minSize;
+
+        RectF rectToBeDivided = minList.get(minSize-1).getBounds();
+        float oLeft = rectToBeDivided.left;
+        float oRight = rectToBeDivided.right;
+        float oTop = rectToBeDivided.top;
+        float oBottom = rectToBeDivided.bottom;
```
ahmedre, line 199: what's the `o` here for? can we expand it?

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+    List<AyahBounds> previousAyahBoundsList = new ArrayList<>(startingBounds);
+    List<AyahBounds> currentAyahBoundsList = new ArrayList<>(coordinatesData.get(currentSurahAyah));
+
+    highlights.clear();
+    highlights.add(ayahTransition);
+
+    // TODO: reason whether this TypeEvaluator should be anonymous or not
+    animator = ValueAnimator.ofObject(new TypeEvaluator() {
+
+      private void normalizeAyahBoundsList(List<AyahBounds> start, List<AyahBounds> end) {
```
ahmedre, line 185: this function would really benefit from some unit tests

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+        List<AyahBounds> start = (List<AyahBounds>)startObject;
+        List<AyahBounds> end = (List<AyahBounds>)endObject;
+
+        if(start.size() != end.size()) {
+          normalizeAyahBoundsList(start, end);
+        }
+
+        int size = start.size();
+
+        //TODO: instead of creating result every time, modify in-place
```
ahmedre, line 222: this matters too because we're in `onDraw` here potentially

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -130,17 +152,178 @@ public void setNightMode(boolean isNightMode, int textBrightness) {
...
+          float right = startValue.right + (endValue.right - startValue.right) * fraction;
+          float bottom = startValue.bottom + (endValue.bottom - startValue.bottom) * fraction;
+          AyahBounds intermediateBounds = new AyahBounds(0,0, new RectF(left, top, right, bottom));
+          result.add(intermediateBounds);
+        }
+        return result;
+      }
+    }, previousAyahBoundsList, currentAyahBoundsList);
+
+    // TODO: where should this constant live?
```
ahmedre, line 239: if you move a lot of this out to new classes, let's move it there

app/src/main/java/com/quran/labs/androidquran/ui/helpers/HighlightType.java
```diff
@@ -8,20 +8,30 @@
...
+  public static final HighlightType AUDIO =     new HighlightType(2, false, R.color.audio_highlight,     true);
+  public static final HighlightType NOTE =      new HighlightType(3, true,  R.color.note_highlight,      false);
+  public static final HighlightType BOOKMARK =  new HighlightType(4, true,  R.color.bookmark_highlight,  false);
 
   private Long id;
   private boolean multipleHighlightsAllowed;
   private int colorId;
   private Integer color = null;
+  private boolean floatable; 
+  // TODO: replace this with HighlightAnimationConfig
```
ahmedre, line 21: considering your comment on 279 of HighlightingImageView and your comment 27 here, i'd go for making this class and only adding the things we need now - i.e. not future proofing it for everything, but making it easy to add say the type of interpolator there later if we want to change it, etc.

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
```diff
@@ -294,6 +477,28 @@ protected void onSizeChanged(int w, int h, int oldw, int oldh) {
...
+      return true;
+    }
+    int arrowIndex = ayah.indexOf("->");
+    if (arrowIndex == -1) {
+      return false;
+    }
+    String startAyah = ayah.substring(0, arrowIndex);
+    String endAyah = ayah.substring(arrowIndex + 2);
+    if(alreadyHighlighted.contains(startAyah)
+        // TODO: to show or not to show is the question!
```
ahmedre, line 492: as a rule of thumb, i'd vote for - when in doubt, don't animate - or when it's complicated to figure it out, don't

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: have expanded it to original. It refers to the original rectangle that has to be divided

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Well, the `evaluate()` function here, provides new values. The `onAnimationUpdate()` hook lets the user use those values and it runs in the UI thread.

I have noticed during my debugging sessions that the `evaluate()` gets called more times than `onAnimationUpdate()` therefore, they are potentially running independent of each other, which makes sense because `onAnimationUpdate()` will be called in an optimized way (once per render frame) but `evaluate()` can be called every update frame. So, we need not worry about our draws getting janked due to `evaluate()` as they are independent of each other.

Moreover, even if we ignore the independence, we can't ignore the fact that if we modify the result in place then we will have a data race condition between the writer (`evaluate()`) and the reader (`onAnimationUpdate()`). We might have a rare  case in which the `result` is simultaneously accessed by `onAnimationUpdate()` and also written by `evaluate()`

So, I have simply removed the TODO and haven't changed the code.

murtraja:
> sorry akhi - i know i promised this ages ago :( added some comments.
> 
> i also tried it and masha'Allah it works pretty well. one question for you also - how do you feel about making the animation simpler - so for example when going from ayah 3 to 4 or 4 to 5 of surah 'abasa, there's a split (or re-join?) for the animation. thinking maybe the 4-5 one should just dismiss the part on the first line and only move the part of the second line. what do you think about simplifying some of this logic?
> 
> jazakAllah khairan for doing this and please forgive my very very very late review.

Would love to simplify this. Right now, if we are animating from ayah a1 -> a2, the number of rectangles being animated is equal to the max rectangles of a1 & a2. What you are suggesting is that if we are animating from 2 rectangles (ayah 4) to 1 rectangle (ayah 5), we should simply focus on 1 rectangle and somehow dispose of the other one. Generally speaking, if going from number of rectangles r1 -> r2, if r1 > r2, use r2 rectangles for animation. But if  r1 < r2, then proceed like normal, split r1 to increase the count till r2. 

If I have understood your point correctly, then we need to think about these cases:
3 -> 2 (discard first rectangle, animate the next two)
3 -> 1 (discard first two rectangles, animate only the last one)
2 -> 1 (discard first rectangle, animate the next one)

Following this approach will save us a "join" at the destination ayah. This effect is more pronounced in 3 -> 1. Surah Al Fajr, ayah 15 to ayah 16. In this case, the lone rectangle signifying ayah end marker "15" will get expanded to fill the rest of the line making up for ayah 16. 

Let me do the refactoring first, so that it is easier to swap out strategies like this, let's see if I am able to make a flexible system that allows us to easily change the methods that we want to use for animation.

Let me know if I have misunderstood your point, or, I was not able to explain my thoughts properly (in which case, will add some videos to show you)

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Have created a class `AyahTransition` that handles this behaviour. Looking at your comment, I am getting an impression that you wanted me to remove the arrow logic altogether, but unfortunately, since we are using HashMap of Strings, I need some key to signify the transition, and the arrow is helpful in that case.

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Have taken the interpolator out of the HighlightingImageView class

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Have moved this to HighlightAnimationConfig

app/src/main/java/com/quran/labs/androidquran/ui/helpers/HighlightType.java
murtraja: Have implemented HighlightAnimationConfig which is flexible enough

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Yes, this method sounds good. I have done exactly that here. 

Say we are playing ayah `a` right now which is already highlighted by the bookmark highlight. Now, when we will switch to playing ayah `b`, then there will be no animation. Only, when both `a` and `b` aren't highlighted, do we show the animation.

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: Right. So this is me being lazy. All I want to do here is simply get the only element from the Set (I know this because, I have checked the size previously). So instead of doing it the obvious iterator way, I followed this approach.

I have fixed this.

app/src/main/java/com/quran/labs/androidquran/widgets/HighlightingImageView.java
murtraja: I have abstracted this away in a class `HighlightNormalizingStrategy` from which all other strategies can be extended. I see that we have already written some tests for data, model, presenter and util. Should I write the test in the util folder or create a new one?

ahmedre:
wa3laikum alsalam,
Eid Mubarak akhi! may Allah accept from all of us - ameen.

jazakumAllah khairan - yep i think we're on the same page. also, wanted to say that masha'Allah code is looking a lot better now! the first few classes make it easy to test since they're pretty much purely math. some minor points:

- `HighlightingImageView` still has way too much code in it and i want to simplify this (even before your PR to be honest, a lot of that is my fault) - just want to avoid making it worse - can we think about moving some more stuff out of it and call whatever needs to be called from one or two entry points from there instead?

- with respect to highlighting keys

> since we are using HashMap of Strings, I need some key to signify the transition, and the arrow is helpful in that case.

what i am saying is we can change it - i guess you're talking about:

```
private SortedMap<HighlightType, Set<String>> currentHighlights = new TreeMap<>();
```

what if we change that instead to not be a `Set<String>` - let's change it to say a `Set<AyahHighlight>` or something, and your `AyahTransition` class can be a subclass of this type?

- regarding `evaluate` and `onAnimationUpdate`

i can't see anything guaranteeing the thread for `evaluate` will be the main thread, though there's a guarantee that side effects of the animator happen on the ui thread. in either case, given how it's being used, i think making  anew one like you are is fine in this case.

jazakumAllah khairan.
walsalam 3alaikum.

murtraja:
Assalaam o Alaikum,
Eid Mubarak to you as well!

Have added the newly proposed strategy and have activated it.


> `HighlightingImageView` still has way too much code in it and i want to simplify this (even before your PR to be honest, a lot of that is my fault)

Would absolutely love to simplify this. But before that, let me try to do this first:

> what if we change that instead to not be a `Set<String>` - let's change it to say a `Set<AyahHighlight>`

Right, so no matter which class we templatize with the `Set`, we need a `hashCode()` implementation. Getting the hash code of strings is pretty easy, so I am thinking of sticking to the same old keys which we have been using to generate the hash code. Let me know if there is another way.

murtraja:
So after replacing String with AyahHighlight, 
1. member variable `floatableAyahCoordinates` which kept track of animation coordinates got merged with existing coordinate data
2. the `onDraw()` got simplified a bit, now that there is only 1 data source


> `HighlightingImageView` still has way too much code in it and I want to simplify this (even before your PR to be honest, a lot of that is my fault)

@ahmedre Since I am not the first author of `HighlightingImageView`, I would request you to please provide some guidelines that I can follow while refactoring this class.

ahmedre:
wa3laikum alsalam,
jazakumAllah khairan - looking good now akhi jazakumAllah khairan - can we move some of the new inner classes of `HighlightingImageView` out into their own standalone classes at least and then we can merge this? also, which evaluator do you think we should use now that you've written both masha'Allah? pick one and let's remove the other one (or if you want we can merge it and then remove it in another PR just to have it in Git history).

barak Allah feekum.
walsalam 3alaikum.

murtraja:
Assalaam o Alaikum,

Have fixed merge conflicts and extracted out `AyahHighlight`, `SingleAyahHighlight` and `TransitionAyahHighlight` into their own classes. Now only two classes remain. `AnimationUpdateListener` and `OverlayParams`. These classes are using some members from the enclosing class, so I am not sure how will I refactor them.

> also, which evaluator do you think we should use now that you've written both masha'Allah?

I will go with your decision on this one.

Shukran

murtraja:
Any updates?

ahmedre:
merging this now, and feel free to work on any improvements after in sha' Allah - jazakumAllah khairan and sorry this took forever :( may Allah reward you greatly!

murtraja:
Shukran. Maybe I will create separate issues/PRs for the improvements you've pointed out - which are mostly cleaning up HighlightingImageView

ahmedre:
please do, jazakumAllah khairan

nacer80:
My  brothers @murtraja & @ahmedre 
Jazekoum Allaho kheiren 
The highlight animation is very great but I have one suggestion, in case of next Ayah is in two lines   is difficult to follow animation in tow ways in the same  time so I think if the highlight animation starts in the first line then the second.
وجزاكم الله خيرا

PR #1318, opened 2020-04-15
Integrate with the audio update api

ahmedre:
This patch implements a weekly check for audio updates. Upon finding
updates, it figures out which files actually require updating and
removes them. A notification is sent to the person to inform them that
some files were erased. An improved ui can be implemented at some point
in the future.


ahmedre:
note to self - i intentionally ignored one edge case, which is "i am playing audio when the updater runs and deletes files" - aside being an edge case, i found this in the `deleteOnExit` docs:

> Note that on Android, the application lifecycle does not include VM termination, so calling this method will not ensure that files are deleted. Instead, you should use the most appropriate out of:
> ...
> * Use the Unix trick of deleting the file as soon as all readers and writers have opened it. No new readers/writers will be able to access the file, but all existing ones will still have access until the last one closes the file.

so if audio started and is playing back and the file is erased, it'll keep playing. if it hasn't quite loaded yet, it'll fail playback and stop audio playback, requiring the person to press play again.

Hiwaselah:
Project in my computer problems

PR #1320, opened 2020-04-18
Minor fixes and logging for Crashlytics issues

ahmedre:


common/search/src/main/java/com/quran/common/search/ArabicSearcher.kt
```diff
@@ -35,23 +36,40 @@ class ArabicSearcher(private val defaultSearcher: Searcher,
 
     val regexp = ArabicCharacterHelper.generateRegex(originalSearchText)
     val pattern = Pattern.compile(regexp)
-    database.rawQuery(query, arrayOf(searchText))?.use { cursor ->
```
ahmedre, line 38: sadly `.use` won't work because `Cursor` doesn't implement `Closable` until 4.2ish

PR #1326, opened 2020-04-23
Update dependency versions

ozbek:
This sets OkHttp and Retrofit minor versions to dynamic, so that we keep receiving (possible future) security updates.


ahmedre:
jazakAllah khairan - not a massive fan of + but since it's a minor version only it should be ok especially since they no longer develop it and only apply critical fixes etc to it

ozbek:
Yes, this is indeed to automatically get critical fixes.
Wa jazakAllah khairan.

PR #1332, opened 2020-05-01
Kotlinize SuraAyah

ahmedre:


app/src/main/java/com/quran/labs/androidquran/data/SuraAyah.kt
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  companion object {
+    @JvmStatic fun min(a: SuraAyah, b: SuraAyah): SuraAyah {
+      return if (a <= b) a else b
+    }
+
+    @JvmStatic fun max(a: SuraAyah, b: SuraAyah): SuraAyah {
+      return if (a >= b) a else b
+    }
```
ahmedre, line 33: we can remove these once the callers are Kotlinized

app/src/main/java/com/quran/labs/androidquran/data/SuraAyah.kt
```diff
@@ -0,0 +1,35 @@
+package com.quran.labs.androidquran.data
+
+import java.io.Serializable
+
+data class SuraAyah(
+  @JvmField val sura: Int,
+  @JvmField val ayah: Int
+) : Comparable<SuraAyah>, Serializable {
```
ahmedre, line 8: Serializable so it can be moved into a jvm only module

PR #1340, opened 2020-05-07
Update French App translation

sbou88:


ahmedre:
jazakumAllah khairan! may you also translate some of the other strings that aren't translated yet if you get a chance?

sbou88:
wa yaak, 
Yes, I can translate missing strings. 
Salam a3laykoum.

PR #1341, opened 2020-05-07
Clean untranslatable

sbou88:
Clean untranslatable string from string.xml


ahmedre:
jazakumAllah khairan - probably should move these out of the main strings.xml one day

PR #1347, opened 2020-05-08
Fix injection of QuranDataSource

ahmedre:


ahmedre:
i should move the provides of `QuranDataSource` at some point in the future since it really doesn't need to be duplicated like this.

PR #1355, opened 2020-05-14
Move 'California' inside parentheses

ozbek:
Because it is not part of the qari's name.


nacer80:
Jazek Allaho khieren my brother Ozbek
You're right good job.

ahmedre:
jazakAllah khairan!

Issue #1361, opened 2020-05-19
expose translation in suralist in a setting


ahmedre:
today, we have a `show_sura_names_translation` boolean that determines whether or not to show the sura name translation in the list screen or not. let's add a preference for this and let people decide (at least for languages where we do have translations of the sura names).

PR #1363, opened 2020-05-20
#1361 show the translated names of the sura in the sura list fragment.

AhmedGarhy:
#1361 show the translated names of the sura in the sura list fragment.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/SuraListFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/SuraListFragment.java
@@ -8,6 +8,7 @@ import androidx.fragment.app.Fragment;
 import androidx.recyclerview.widget.DefaultItemAnimator;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -28,6 +29,7 @@ import javax.inject.Inject;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.observers.DisposableSingleObserver;
+
 import org.jetbrains.annotations.NotNull;
 
 import static com.quran.labs.androidquran.data.Constants.JUZ2_COUNT;
@@ -46,6 +48,8 @@ public class SuraListFragment extends Fragment {
 
   @Inject QuranInfo quranInfo;
   @Inject QuranDisplayData quranDisplayData;
+  @Inject QuranSettings quranSettings;
+  boolean showSuraTranslatedName = false;
   private int numberOfPages;
 
   public static SuraListFragment newInstance() {
@@ -88,8 +92,11 @@ public class SuraListFragment extends Fragment {
   public void onResume() {
     final Activity activity = getActivity();
     if (activity instanceof QuranActivity) {
-      QuranSettings settings = QuranSettings.getInstance(activity);
-      ((QuranListAdapter) mRecyclerView.getAdapter()).showHideSuraTranslatedName(settings.isShowSuraTranslatedName());
+      boolean newValueOfShowSuraTranslatedName = quranSettings.isShowSuraTranslatedName();
+      if (this.showSuraTranslatedName != newValueOfShowSuraTranslatedName) {
+        showHideSuraTranslatedName(newValueOfShowSuraTranslatedName);
+        this.showSuraTranslatedName = newValueOfShowSuraTranslatedName;
+      }
       disposable = ((QuranActivity) activity).getLatestPageObservable()
           .first(Constants.NO_PAGE)
           .observeOn(AndroidSchedulers.mainThread())
@@ -109,7 +116,7 @@ public class SuraListFragment extends Fragment {
             }
           });
 
-      if (settings.isArabicNames()) {
+      if (quranSettings.isArabicNames()) {
         updateScrollBarPositionHoneycomb();
       }
     }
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/fragment/SuraListFragment.java
```diff
@@ -128,7 +129,8 @@ private void updateScrollBarPositionHoneycomb() {
 
     Activity activity = getActivity();
     boolean wantPrefix = activity.getResources().getBoolean(R.bool.show_surat_prefix);
-    boolean wantTranslation = activity.getResources().getBoolean(R.bool.show_sura_names_translation);
+    QuranSettings settings = QuranSettings.getInstance(activity);
```
ahmedre, line 132: replace with:

```
  @Inject
  QuranSettings quranSettings;
```

field at the top.

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranListAdapter.java
```diff
@@ -248,6 +249,35 @@ public boolean onLongClick(View v) {
...
+
+  private void addTranslatedNameToSuraName(QuranRow row) {
+    StringBuilder builder = new StringBuilder(row.text);
+    String translation = context.getResources().getStringArray(R.array.sura_names_translation)[row.sura - 1];
+    if (!TextUtils.isEmpty(translation)) {
+      builder.append(" (" + translation + ")");
+    }
+    row.text = builder.toString();
+  }
+
```
ahmedre, line 280: suggestion - instead of doing it this way, i would remove this code and let changing the variable cause `SuraListFragment` instead to recompute with the updated flag and `notifyDataSetChanged`.

app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -167,6 +167,9 @@ public void setShowDate(boolean isDateShown) {
     prefs.edit().putBoolean(Constants.PREF_SHOW_DATE, isDateShown).apply();
   }
 
+  public boolean isShowSuraTranslatedName() {
+    return prefs.getBoolean(Constants.PREF_SURA_TRANSLATED_NAME, false);
```
ahmedre, line 171: can you default to the value in `R.bool.show_sura_names_translation`

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranListAdapter.java
ahmedre: so you could replace that code with a check in `onResume` - if the value of the checkbox changed, then re-compute and `notifyDataSetChanged`, otherwise do nothing.

ahmedre:
jazakAllah khairan for taking this up! just a few minor comments.

app/src/main/java/com/quran/labs/androidquran/ui/helpers/QuranListAdapter.java
AhmedGarhy: okay, I will move this code [three methods] to the SuraListFragment, but I track if the value changed or not using the below-stopping conditions before re-compute. 

 if (showSuraTranslatedName && elements[1].text.indexOf("(") != -1) return;
 if (!showSuraTranslatedName && elements[1].text.indexOf("(") == -1) return;

- do you know how to track the value after being changed ?

ahmedre: no need to move this code - remove it altogether. in `onCreate` of the `SuraListFragment`, read the value of `wantTranslations` into a variable. in `onResume`, check if the actual value differs, and if so, just re-call `getSuraList()` - pass the updated value to the adapter (you might have to expose a setter).

app/src/main/java/com/quran/labs/androidquran/ui/fragment/SuraListFragment.java
```diff
@@ -88,7 +92,11 @@ public void onPause() {
   public void onResume() {
     final Activity activity = getActivity();
     if (activity instanceof QuranActivity) {
-      QuranSettings settings = QuranSettings.getInstance(activity);
+      boolean newValueOfShowSuraTranslatedName = quranSettings.isShowSuraTranslatedName();
+      if (this.showSuraTranslatedName != newValueOfShowSuraTranslatedName) {
+        showHideSuraTranslatedName(newValueOfShowSuraTranslatedName);
+        this.showSuraTranslatedName = newValueOfShowSuraTranslatedName;
```
ahmedre, line 98: what i meant is replace these 2 lines with something like:

```java
final QuranRow[] updatedElements = getSuraList();
adapter.setElements(updatedElements);
```

and in `onCreate`, make the adapter in a variable so you can call this setter on it. expose the setter on the adapter also. then you can just remove the show/add methods that you added below.

ahmedre:
sorry for the delayed reply and jazakAllah khairan for your efforts - just the minor comment in sha' Allah after that can merge. barak Allah feek.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/SuraListFragment.java
AhmedGarhy: Hi @ahmedre 
I got what do you want to say, but I believe that we don't have to repeat the construction of the SuraList by Calling getSuraList() again, instead of that, we can just modify the text of each Sura in the list by appending or removing the translated part.

I believe this implementation is more optimal than calling the getSuraList again and again.

ahmedre: i agree that it's more optimal your way, but considering:

1. we can avoid adding extra code - so tradeoff for code readability
2. the extra price is only called when the person goes into the menu and chooses the option to show translation names - the number of people doing this is very small and will likely only do it once or twice at most.
3. the extra code means that in the future, if we change the way we're adding the translations (i.e. today, i guess we're manually making the string, which is not ideal and honestly we should fix this), then we'd have to modify this code - it also means other languages can't choose to something different - ex one language can use () one can use <>, one can do nothing at all.

so for those reasons, was suggesting to do it this way despite being slightly less optimal.

AhmedGarhy: got it, thanks for the clarification.

ahmedre:
jazakAllah khairan!

PR #1371, opened 2020-05-31
1365/ Add ToastCompat with support for dark toasts on pre-R

ataulm:
I started implementing this using theme attributes, but then I noticed there was an in-app night mode. I tried to follow this approach which means that the system "Night mode" will have no effect on the color of the toast.

Before / Default | Night mode (in-app) enabled
---|---
lightmode | nightmode

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/QuranAdvancedPreferenceActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/QuranAdvancedPreferenceActivity.java
@@ -16,6 +16,7 @@ import android.widget.Toast;
 import com.quran.labs.androidquran.service.util.PermissionUtil;
 import com.quran.labs.androidquran.ui.QuranActionBarActivity;
 import com.quran.labs.androidquran.ui.fragment.QuranAdvancedSettingsFragment;
+import com.quran.labs.androidquran.ui.util.ToastCompat;
 import com.quran.labs.androidquran.util.AudioManagerUtils;
 import com.quran.labs.androidquran.util.QuranSettings;
 
@@ -82,7 +83,7 @@ public class QuranAdvancedPreferenceActivity extends QuranActionBarActivity {
           REQUEST_WRITE_TO_SDCARD_PERMISSION);
     } else {
       // in the future, we should make this a direct link - perhaps using a Snackbar.
-      Toast.makeText(this, R.string.please_grant_permissions, Toast.LENGTH_SHORT).show();
+      ToastCompat.makeText(this, R.string.please_grant_permissions, Toast.LENGTH_SHORT).show();
     }
   }
 
--- a/app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
@@ -46,7 +46,7 @@ public class ToastCompat {
       toastView.setBackgroundDrawable(tintedBackground);
     }
 
-    TextView textView = toastView.findViewById(android.R.id.message);
+    TextView textView = toastView.findViewById(R.id.message);
     textView.setTextColor(getTextColor(context, nightMode));
     textView.setText(message);
     return toastView;
--- a/app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
@@ -1,0 +1,83 @@
+package com.quran.labs.androidquran.ui.util;
+
+import android.annotation.SuppressLint;
+import android.content.Context;
+import android.graphics.Color;
+import android.graphics.drawable.Drawable;
+import android.os.Build;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.TextView;
+import android.widget.Toast;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.StringRes;
+import androidx.appcompat.content.res.AppCompatResources;
+import androidx.core.content.ContextCompat;
+import androidx.core.graphics.drawable.DrawableCompat;
+
+import com.quran.labs.androidquran.R;
+import com.quran.labs.androidquran.util.QuranSettings;
+
+public class ToastCompat {
+
+  private static final int ANDROID_R = 30;
+
+  public static Toast makeText(@NonNull Context context, @StringRes int messageRes, int duration) {
+    return makeText(context, context.getString(messageRes), duration);
+  }
+
+  public static Toast makeText(@NonNull Context context, CharSequence message, int duration) {
+    @SuppressLint("ShowToast") Toast toast = Toast.makeText(context, message, duration);
+    if (Build.VERSION.SDK_INT < ANDROID_R) {
+      // It's unclear what behavior `Toast.setView()` has on R and above, hence the guard.
+      toast.setView(createToastView(context, message));
+    }
+    return toast;
+  }
+
+  private static View createToastView(Context context, CharSequence message) {
+    View toastView = LayoutInflater.from(context).inflate(R.layout.toast, null, false);
+
+    QuranSettings settings = QuranSettings.getInstance(context);
+    Drawable tintedBackground = createTintedBackground(context, settings.isNightMode());
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+      toastView.setBackground(tintedBackground);
+    } else {
+      toastView.setBackgroundDrawable(tintedBackground);
+    }
+
+    TextView textView = toastView.findViewById(android.R.id.message);
+    textView.setTextColor(getTextColor(context, settings));
+    textView.setText(message);
+    return toastView;
+  }
+
+  private static Drawable createTintedBackground(Context context, boolean nightMode) {
+    Drawable toastFrame = DrawableCompat.wrap(AppCompatResources.getDrawable(context, R.drawable.toast_frame));
+    DrawableCompat.setTint(toastFrame, getBackgroundColor(context, nightMode));
+    return toastFrame;
+  }
+
+  private static int getTextColor(Context context, QuranSettings settings) {
+    if (settings.isNightMode()) {
+      int textColorNight = ContextCompat.getColor(context, R.color.toast_text_color_night);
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/quran/labs/androidquran/QuranAdvancedPreferenceActivity.java
```diff
@@ -82,7 +83,7 @@ public void requestWriteExternalSdcardPermission(String newLocation) {
           REQUEST_WRITE_TO_SDCARD_PERMISSION);
     } else {
       // in the future, we should make this a direct link - perhaps using a Snackbar.
-      Toast.makeText(this, R.string.please_grant_permissions, Toast.LENGTH_SHORT).show();
+      ToastCompat.makeText(this, R.string.please_grant_permissions, Toast.LENGTH_SHORT).show();
```
ataulm, line 86: It's not possible for us to change the colors without getting a reference to the Toast itself because:

1) the attribute it uses for the background is private
2) it doesn't use attributes for the text color - it's hardcoded

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+import androidx.appcompat.content.res.AppCompatResources;
+import androidx.core.content.ContextCompat;
+import androidx.core.graphics.drawable.DrawableCompat;
+
+import com.quran.labs.androidquran.R;
+import com.quran.labs.androidquran.util.QuranSettings;
+
+public class ToastCompat {
+
+  private static final int ANDROID_R = 30;
```
ataulm, line 24: There's no built-in constant for this (potentially because the target SDK of this app is 28).

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public static Toast makeText(@NonNull Context context, @StringRes int messageRes, int duration) {
+    return makeText(context, context.getString(messageRes), duration);
+  }
+
+  public static Toast makeText(@NonNull Context context, CharSequence message, int duration) {
+    @SuppressLint("ShowToast") Toast toast = Toast.makeText(context, message, duration);
+    if (Build.VERSION.SDK_INT < ANDROID_R) {
+      // It's unclear what behavior `Toast.setView()` has on R and above, hence the guard.
+      toast.setView(createThemeFriendlyToastView(context, message));
```
ataulm: I suspect that this would still continue to work on Android R (`Toast.getView()`/`Toast.setView(View)` is deprecated from that version and I read that `getView()` will return `null`) but I haven't been able to test it. 

When it can be tested/when the behaviour will be clearer, this guard could be removed/amended.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public static Toast makeText(@NonNull Context context, @StringRes int messageRes, int duration) {
+    return makeText(context, context.getString(messageRes), duration);
+  }
+
+  public static Toast makeText(@NonNull Context context, CharSequence message, int duration) {
+    @SuppressLint("ShowToast") Toast toast = Toast.makeText(context, message, duration);
+    if (Build.VERSION.SDK_INT < ANDROID_R) {
+      // It's unclear what behavior `Toast.setView()` has on R and above, hence the guard.
+      toast.setView(createToastView(context, message));
```
ataulm, line 34: `Toast.getView()` and `Toast.setView()` are deprecated in Android R. I read (somewhere) that `getView()` will return `null`, and I don't know what the behaviour of `setView()` will be, though I wouldn't be surprised if it's a no-op on Android R (if `targetSdkVersion >= 30`).

I wasn't able to test on R, so I left this with a guard. This means that on R and above, unless this logic is amended, toasts will do whatever they do by default, which means that the toast will be light colored.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private static View createToastView(Context context, CharSequence message) {
+    View toastView = LayoutInflater.from(context).inflate(R.layout.toast, null, false);
+
+    QuranSettings settings = QuranSettings.getInstance(context);
+    Drawable tintedBackground = createTintedBackground(context, settings.isNightMode());
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
+      toastView.setBackground(tintedBackground);
+    } else {
+      toastView.setBackgroundDrawable(tintedBackground);
```
ataulm, line 47: ⚠️  I have not tested this on anything apart from API 29.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    TextView textView = toastView.findViewById(android.R.id.message);
```
ataulm, line 50: I continued to use the same IDs that the default Toast layout used.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private static int getTextColor(Context context, QuranSettings settings) {
+    if (settings.isNightMode()) {
+      int textColorNight = ContextCompat.getColor(context, R.color.toast_text_color_night);
+      return Color.argb(
+          settings.getNightModeTextBrightness(),
+          Color.red(textColorNight),
+          Color.green(textColorNight),
+          Color.blue(textColorNight)
```
ataulm, line 69: I've seen `getNightModeTextBrightness()` used elsewhere, but with `255` hardcoded. Is that preferred?

app/src/main/res/drawable/toast_frame.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+  <!-- background is tinted programmatically -->
+  <solid android:color="@android:color/holo_red_light" />
```
ataulm, line 5: This should always be tinted - if the toast looks red, something is not correct 😅

ataulm:
FWIW, I don't know whether it's worth merging this/trying to implement this feature.

From my quick test last week, it looks like the default behaviour on Android, even on Android 10 with night mode enabled, is for Toasts not to change appearance. This + the deprecated API on Android 11 suggests that it's not worth the effort.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ahmedre: let's play it safe here and do nothing if it's null just in case Samsung devices do something odd

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ahmedre: if `Toast.getView()` is not deprecated until R and we're only supporting this R and below, could we not just let the normal Toast ui inflate itself, get its view, and if it's not null, apply a background instead?

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ahmedre: i think for the alpha, it's fine to hardcode to 255 here since we don't want the background to be very transparent and perhaps not very readable? as a matter of fact i am ok with adding a toast_text_color_night (until we support day/night) and just using a nice value for that instead of making it customizable

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ataulm: Yep we can, but as you mentioned, different device manufacturers could do odd things so we'd have to set the background and text color in any case.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ataulm: This will never be null as we inflate a layout `R.layout.toast` which uses this ID. I can rename the ID?

I wasn't sure if the system would rely on this ID being something particular, but it shouldn't.

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ataulm: This function is for the text color so it will only affect the contrast ratio of the text against the background by modifying the text alpha: the background of the toast will not change transparency.

I'm not sure what the requested action is on this one - are we saying to get rid of the `getNightModeTextBrightness` like this:

```java
private static int getTextColor(Context context) {
  if (settings.isNightMode()) {
    return ContextCompat.getColor(context, R.color.toast_text_color_night);
  } else {
    return ContextCompat.getColor(context, R.color.toast_text_color);
  }
}
```

or something else?

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ahmedre: ah right you're inflating your own layout - never mind then, this is fine

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ahmedre: yes exactly - handle the same way you're handling day mode where we just provide a specific background color and text color

app/src/main/java/com/quran/labs/androidquran/ui/util/ToastCompat.java
ataulm: *we can but I don't recommend. If the device manufacturer inflates its own layout for example, we wouldn't know the ID for the textview.

PR #1386, opened 2020-06-27
Fix some bugs with AudioUtils

ahmedre:


app/src/main/java/com/quran/labs/androidquran/presenter/audio/AudioPresenter.kt
```diff
@@ -50,13 +50,8 @@ constructor(private val quranDisplayData: QuranDisplayData,
         audioPathInfo
       }
 
-      val (checkedStart, checkedEnd) = if (start < end) start to end else end to start
-      if (checkedStart != start) {
-        Crashlytics.logException(IllegalArgumentException("expected $start > $end, but wasn't."))
-      }
```
ahmedre, line 56: never actually happened

app/src/main/java/com/quran/labs/androidquran/util/AudioUtils.kt
```diff
@@ -125,19 +125,7 @@ constructor(private val quranInfo: QuranInfo, private val quranFileUtils: QuranF
...
-      // using [page+1] as an index because we literally want the next page
-      val nextPageAyah = quranInfo.getFirstAyahOnPage(nextPage)
-
-      pageLastSura = nextPageSura
-      pageLastAyah = nextPageAyah - 1
-      if (pageLastAyah < 1) {
-        pageLastSura--
-        pageLastAyah = quranInfo.getNumberOfAyahs(pageLastSura)
-      }
-    }
```
ahmedre, line 140: this was for page download amount, and there's a method for that which is also more correct than this (this code is actually wrong if you consider a case like 56:51, which was returning the sura as 57 instead of 56)

app/src/test/java/com/quran/labs/androidquran/util/AudioUtilsTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.quran.data.core.QuranInfo
+import com.quran.data.model.SuraAyah
+import com.quran.data.pageinfo.common.MadaniDataSource
+import com.quran.data.source.PageProvider
+import org.junit.Assert
+import org.junit.Test
+import org.mockito.Mockito
+import org.mockito.Mockito.`when` as whenever
+
+class AudioUtilsTest {
```
ahmedre, line 13: converted to Kotlin

app/src/test/java/com/quran/labs/androidquran/util/AudioUtilsTest.kt
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Test
+  fun testLastAyahForFirstAyahWithPageDownload() {
+    val audioUtils = AudioUtils(QuranInfo(MadaniDataSource()),
+        Mockito.mock(QuranFileUtils::class.java))
+    val start = SuraAyah(56, 51)
+    val end = audioUtils.getLastAyahToPlay(start, 536, 1, false)
+    assertThat(end).isEqualTo(SuraAyah(56, 76))
+  }
```
ahmedre, line 59: added this test - was failing before my code change

PR #1387, opened 2020-06-27
Kotlinize QuranFileUtils

ahmedre:


app/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
ahmedre, line 1: filed #1388 to remove this in the future

PR #1389, opened 2020-06-27
Initial handling of local data

ahmedre:
For warsh, the plan is to replace the set of images with an updated set.
To make life easier, the plan is to bundle these with the apk. This
patch allows for copying bundled data for sets that support it instead
of downloading that data from a server.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.kt
@@ -7,6 +7,7 @@ import android.graphics.Bitmap.Config.ALPHA_8
 import android.graphics.BitmapFactory
 import android.graphics.BitmapFactory.Options
 import android.os.Environment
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.kt
```diff
@@ -195,6 +220,25 @@ class QuranFileUtils @Inject constructor(
...
+        }
+
+    if (filename.endsWith(".zip")) {
+      val zipFile = destination + File.separator + filename
+      ZipUtils.unzipFile(zipFile, destination, filename, null)
+      // delete the zip file, since there's no need to have it twice
+      File(zipFile).delete()
+    }
+  }
+
```
ahmedre, line 241: `CopyDatabaseUtil` has similar code but will be removed in a future PR in sha' Allah since it's not used at the present.

PR #1394, opened 2020-07-05
Support split screen age and translation on tablet

### LINKED ISSUE #759: Support split screen page and translation on tablet
A lot of this work is actually already done - the pieces left are:

1. in `TabletFragment`'s `getAyahTrackerItems`, when the mode is translation, return an arabic tracker and a translation tracker.

2. in `TabletFragment`'s `onCreateView`, call `init` with the proper page types - i.e. with `TabletView.QURAN_PAGE, TabletView.TRANSLATION_PAGE` (if the Quran page is on the right, or the opposite if the Quran page is on the left).

3. expose a switch to enable this somewhere in the ui - let's discuss ideas about this on the ticket - the very simplest would be a setting in the settings menu, which would work, but wouldn't be ideal.

4. the adapter needs to know that mixed mode is 604 pages, not 302.

miscellaneous points:

- when the Quran page is on the left, let the Quran page be on the left and the translation on the right, and vice versa.

- when refactoring the adapter, let's think about cleaning it up - maybe injecting different types of adapters that just override whatever one or two methods are extra or different instead of all the nested if/elses.

cc @ozbek and @hosainnet since one of you maybe interested in doing this - please comment before starting just to avoid duplication of work.

Korogba:
This PR Fixes #759.

It allows a second translation mode with the Quran and translation placed side-by-side on the same page. 

This mode can be enabled in the settings. Also, the placement of the Quran/translation can be configured in settings.

Before merging this PR, approval and translations will be needed for the strings shown in the settings page.


app/src/main/java/com/quran/labs/androidquran/util/AudioUtils.kt
```diff
@@ -107,17 +107,25 @@ constructor(private val quranInfo: QuranInfo, private val quranFileUtils: QuranF
...
+  fun getLastAyahToPlay(
+    startAyah: SuraAyah,
+    currentPage: Int,
+    mode: Int,
+    isDualPages: Boolean,
+    isNotSplitOrShowingTranslation: Boolean
+  ): SuraAyah? {
+    val page =
+      if (isDualPages &&
+        isNotSplitOrShowingTranslation &&
```
ahmedre, line 119: this isn't right for cases where we are tablet in translation mode without the split setting - in that case, we should still treat it as 2 pages whereas this code will treat it as 1.

common/data/src/main/java/com/quran/data/core/QuranInfo.kt
```diff
@@ -160,16 +160,26 @@ class QuranInfo @Inject constructor(quranDataSource: QuranDataSource) {
     return if (sura < 1 || sura > NUMBER_OF_SURAS) -1 else suraNumAyahs[sura - 1]
   }
 
-  fun getPageFromPosition(position: Int, dual: Boolean): Int {
-    return if (dual) {
+  fun getPageFromPosition(
+    position: Int,
+    dual: Boolean,
+    isSplitScreen: Boolean,
+    isShowingTranslation: Boolean
```
ahmedre, line 167: same comment as above

common/data/src/main/java/com/quran/data/core/QuranInfo.kt
```diff
@@ -160,16 +160,26 @@ class QuranInfo @Inject constructor(quranDataSource: QuranDataSource) {
...
     }
   }
 
-  fun getPositionFromPage(page: Int, dual: Boolean): Int {
-    return if (dual) {
+  fun getPositionFromPage(
+    page: Int,
+    dual: Boolean,
+    isSplitScreen: Boolean,
+    isShowingTranslation: Boolean
```
ahmedre, line 180: here also

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -203,6 +202,7 @@
   private int numberOfPages;
   private int numberOfPagesDual;
   private int defaultNavigationBarColor;
+  private boolean isSplitScreen = false;
```
ahmedre, line 205: can we unify some code by making this an int - numberOfActualPagesVisible for example that's set to 1 for single page (or dual page with split turned on) and is 2 for dual mode (irrespective of translation on or off?)

ahmedre:
2 questions about this - curious as to both your thoughts @Korogba @ozbek 

1. right now there's this option for "Quran page on the right side" that controls whether the translation is always on the left and the Quran page is always on the right or the other way around. i am not too sure about this option - my initial thought is maybe not to have this at all and always match the Arabic page to where it would be in the mushaf (i.e. page 1 would always be on the right and page 2 would always be on the left and so on) - my concern though is maybe this would be too much moving around to be comfortable for reading for people, but not sure. what do you think about this? if you think there's still value of having a manual setting for right vs left, does it make sure to have this setting as opposed to just deciding one and keeping that one?

2. regarding the settings - it's confusing that the dual screen option itself is under display settings, and the rest of these are under translations (though to be honest, i understand it and it makes sense). maybe it would make sense to have a section for dual screen and translation settings or something that has both sets of settings, but not sure that would make it better. open to more thoughts about this.

jazakumAllah khairan and please forgive my lateness in reviewing this!

app/src/main/java/com/quran/labs/androidquran/util/AudioUtils.kt
Korogba: Will have a look this weekend Insha Allah. Same for all the other comments.

Korogba:
@ahmedre Wa ʾantum fa-jazakumu-Allahu khayran. Thank you for reviewing. 

1. I am not sure about matching the page as it will be in the mushaf. For one, it might seem inconsistent to the reader if the translation/Arabic page keeps switching as they read along. I generally agree with deciding on a placement and just sticking to that.

2. What about having a separate section for dual screen like you mentioned. That will have the toggle to turn on/off dual mode and other settings related to dual mode as introduced in this PR.

ahmedre:
sounds good - let's keep 1 as is for now and consider later. and for 2, yeah a separate section for dual page settings

app/src/main/java/com/quran/labs/androidquran/util/AudioUtils.kt
Korogba: Salaam walaikum, I finally got around to this 😅. 

I changed this function along with others with a similar signature to depend on a single boolean that determines if two/one page(s) are/is displayed - 702469884a9d289c237c974edda1aa66bceaf896.

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
Korogba: I left this boolean because it is useful to save the preferences value once but the usage is modified so we now have a function that determines if dual page or not. Since this value changes each time the user turns translation on/off, most of the use for this boolean is replaced by a function - `isDualPageVisible()`. See 702469884a9d289c237c974edda1aa66bceaf896 and 70544ae4d691e9e341872170fa5dca4878c9d2c7.

Korogba:
> sounds good - let's keep 1 as is for now and consider later. and for 2, yeah a separate section for dual page settings

Added separate category for dual page: 9552d020cd4d7ca117149e3fd208406022cf23fc.

ahmedre:
jazakumAllah khairan - merged it

PR #1397, opened 2020-07-06
Fix audio not refreshing for the last ayah

ahmedre:
In certain cases, the audio stopped refreshing for the last ayah. This
isn't right, since in some cases, we have a 999 stopping marker that
should be stopped at (instead of the end of the file). This just checks
the position periodically, even when we're at the last ayah.


ahmedre:
you can try sheikh Mishari last ayah of sura Kahf without and with this patch (assuming you have the latest database version). with this patch, you don't get the copyright speech after the end of the ayah. without it, you do.

PR #1421, opened 2020-08-17
Update Kotlin to 1.4 and update dependencies

ahmedre:


app/build.gradle
```diff
@@ -130,7 +130,6 @@ dependencies {
 
   implementation project(path: ':feature:audio')
 
-  implementation deps.kotlin.stdlib
```
ozbek, line 133: Yay, one less dependency! )

ahmedre: still needed but it's transparently pulled in now by the Kotlin plugin

PR #1428, opened 2020-09-02
Change right/left buttons glyph orientation to up/down

amynbe:
السلام عليكم

I always get confused by the navigation buttons in the translation panel.

The convention we're used to, in left-to-right apps and books, is left=back and right=forward.
Although we know that Arabic is written right to left, the translation is often in a Latin language, leaving the user's confused as to which convention to work with.

Over the years, I never managed to get rid of this confusion and still find myself trying both ways without really caring about the button glyphs.

An up-down convention seems to make more sense, especially since the previous/next ayah is often located just **above** or **under** the current one, with the top and bottom ayah of each page being edge cases.

What do you think @ahmedre @ozbek ? 

prev-next-btn-orientation-2


nacer80:
وعليكم السلام ورحمة الله وبركاته 
You are right my brother، it looks very helpful.
جزاك الله خيرا

benomaire:
I feel for you! However, I feel the current RTL way is more natural if some arabic tafsir is shown.

May be your propsed solution could make another confusion if many translations are activated, since there will be two up/down motions: the swipe for current ayah, and buttons for changing the ayah.

I do not oppose your proposal, but here are other ideas to think about:
1.  Just change the buttons behaviour based on UI language (by default), may be an option is offered. In this case only if the UI is arabic, the current RTL convention is used.
2. Stack the buttons on top of each other, at least the user can use one hand easily. May be it would be useful to stack the surah number and ayah number as well to give immediate feedback and avoid possible up/down confusion, although this will cover too much text. See the screenshots:
<details><summary>Screenshots</summary>




</details>

amynbe:
@ozbek @nacer80 @benomaire Thanks brothers for the feedback.

@benomaire :
1- I think the button confusion comes more from the fact that you have an Arabic mus'haf mixed with Latin translation, regardless of UI language. You can see in the screenshots very few UI text compared to the amount of actual content. So changing the default based on UI language I expect will only add more confusion and not solve the problem.

2- I personally completely agree with the idea of buttons on the same side of the screen as I also struggle to press both with one thumb. But I think that can be kept as a separate, maybe slightly less important improvement than the cognitive problem discussed here.

nacer80:
My brothers 
It's good idea to make the buttons one same side and same row like the second screen shot.
وجزاكم الله خيرا وتقبل الله منا ومنكم

nacer80:
I think if we make it on the center is batter
As shown in the picture below:

sempsmahad:
I would go with the centred one too if the buttons where a bit bigger and the surah number dint look like a button

amynbe:
Not sure if you got notification for this, but see new PR "Center prev next buttons" #1431 for the discussed change.

ahmedre (approved):
jazakumAllah khairan, this was always confusing to me too!

Issue #1407, opened 2020-07-22
Improvements to audio settings


ahmedre:
From an email (and I agree with these making sense):

> 1. When I play a Surat then I go into settings to chose to play from what Surat to what Surat and I click play only above verses, the audio of the Surat will restart. It really shouldn't have to restart and should just apply the settings and keep playing. It makes it harder for the kids to focus because it repeats when I apply the settings.

> 2. There should be an option to remember the last selected setting being "play only above verses". I'm always having to manually select it.

nacer80:
السلام عليكم ورحمة الله وبركاته 
I got this comment from Google play store.

**8/25/20  

Assalamualaikum. In this new version, I cant select the number of ayahs to be played. Once we select the 'play' button, it will play the entire surah. The previous version let us select the number of ayahs to be played.**

arman2123:
Is this issue solved? Any further improvement needed on this area?

sempsmahad:
@arman2123 not yet solved @ahmedre had told me to make some changes to my solution

ahmedre:
وعليكم السلام ورحمة الله وبركاته
this functionality never went away - start playing - at the bottom of the screen in the audio bar is a settings icon. tap it and you can select the number of verses being played, how many to repeat, and so on.
والسلام عليكم

sempsmahad:
وعليكم السلام ورحمة الله وبركاته
@ahmedre i think i will have something by monday next week إِنْ شَاءَ ٱللَّٰ

arman2123:
I am new to this project, although I use quran.com regularly. I'll try to get the working in the local dev environment. I'll get back to the github Issues later to pick up tasks when I can handle them. 

Do you guys have any channel where you communicate? Slack, teams etc. Just in case if I need to discuss something.

reinaldoriant:
@ahmedre 
السلام عليكم ورحمة الله وبركاته
I tried this repro, but I didn't find yet the problem in the first opinion. But the second opinion is maybe I can help using data store, can I?.

shahawi:
السلام عليكم و رحمة الله و بركاته

I was wondering what is the state of this issue, And how can I help instead of starting over.

ahmedre:
وعليكم السلام ورحمة الله وبركاته
I believe this was fixed and merged long ago.
please open a new issue if there are still problems with it.
would be glad to have your contribution on any improvements to the code though, please feel free to open a ticket with some info about what you want to do / improve before building it.
[...]

PR #1429, opened 2020-09-05
#1407 improvements to audio settings

sempsmahad:
1.When I play a Surat then I go into settings to chose to play from what Surat to what Surat and I click play only above verses, the audio of the Surat will restart. It really shouldn't have to restart and should just apply the settings and keep playing. It makes it harder for the kids to focus because it repeats when I apply the settings.

A. removed the part where audio playback is stopped if playing now it pauses and continues on pressing play if in selected range restarts to new start ayah if not in selected range

2.There should be an option to remember the last selected setting being "play only above verses". I'm always having to manually select it.

A.modified 'refreshview()' to read from former used settings which where stored in sharedprefferences on pressing apply button

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -172,6 +172,7 @@ dependencies {
   androidTestImplementation "androidx.test.espresso:espresso-core:${espressoVersion}"
   androidTestImplementation "androidx.test.espresso:espresso-intents:${espressoVersion}"
   implementation 'com.crashlytics.sdk.android:crashlytics:2.10.1'
+  implementation 'com.google.code.gson:gson:2.8.6'
 
   errorprone 'com.google.errorprone:error_prone_core:2.3.3'
   //noinspection GradleDynamicVersion
--- a/app/src/main/java/com/quran/labs/androidquran/dao/audio/AudioPlaybackSettings.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/dao/audio/AudioPlaybackSettings.kt
@@ -1,0 +1,12 @@
+package com.quran.labs.androidquran.dao.audio
+
+import android.os.Parcelable
+import com.quran.data.model.SuraAyah
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class AudioPlaybackSettings(val start: SuraAyah,
+                                 val end: SuraAyah,
+                                 val verseRepeatCount: Int = 0,
+                                 val rangeRepeatCount: Int = 0,
+                                 val enforceRange: Boolean = false) : Parcelable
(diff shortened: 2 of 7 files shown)


app/build.gradle
```diff
@@ -172,6 +172,7 @@ dependencies {
   androidTestImplementation "androidx.test.espresso:espresso-core:${espressoVersion}"
   androidTestImplementation "androidx.test.espresso:espresso-intents:${espressoVersion}"
   implementation 'com.crashlytics.sdk.android:crashlytics:2.10.1'
+  implementation 'com.google.code.gson:gson:2.8.6'
```
ahmedre, line 175: we already use moshi so let's not add gson

app/src/main/java/com/quran/labs/androidquran/dao/audio/AudioPlaybackSettings.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.os.Parcelable
+import com.quran.data.model.SuraAyah
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class AudioPlaybackSettings(val start: SuraAyah,
+                                 val end: SuraAyah,
+                                 val verseRepeatCount: Int = 0,
+                                 val rangeRepeatCount: Int = 0,
+                                 val enforceRange: Boolean = false) : Parcelable
```
ahmedre, line 12: what if, instead of doing this, we remove this class (and the corresponding changes to read/write this) - and instead do the work in the `AudioService`?

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
```diff
@@ -414,15 +414,8 @@ private void handleIntent(Intent intent) {
         audioQueue = new AudioQueue(quranInfo, audioRequest,
             new AudioPlaybackInfo(start, 1, 1, basmallah));
         Crashlytics.log("audio request has changed...");
-
-        if (player != null) {
-          player.stop();
-        }
-        state = State.Stopped;
-        Crashlytics.log("stop if playing...");
```
ahmedre, line 422: stopping is sometimes the right thing to do - but this is where we should check. you have the old `audioRequest` (before we override it in line 409). we can check for example if the current verse is within the range of playback (i.e. between `oldAudioRequest.getStart()` and `oldAudioRequest.getEnd()`), then don't stop. otherwise stop, and in either case, we'll replace the `audioRequest`.

ahmedre (commented):
jazakumAllah khairan, this is a much needed change. left some minor comments.

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
sempsmahad: Thanks for the reply @ahmedre I will work on the changes as soon as possible 
Ma'Assalam

PR #1434, opened 2020-09-11
Use red color for downloaded surah in Audio Manager

benomaire:
Suggestion to change the color of the downloaded Surah in Audio Manager.
I feel this is very useful to quickly locate the downloaded Surah's in the list.

redDeleteColor


ahmedre:
جزاكم الله خيراً

app/src/main/java/com/quran/labs/androidquran/ui/SheikhAudioManagerActivity.kt
```diff
@@ -374,15 +374,19 @@ class SheikhAudioManagerActivity : QuranActionBarActivity(), SimpleDownloadListe
       holder.name.text = quranDisplayData.getSuraName(context, position + 1, true)
       val surahStatus: Int
       val surahStatusImage: Int
+      val surahStatusBackground: Int
```
ahmedre, line 377: could we have just set the color directly as the background?

benomaire: We can but the button will be square. I wanted to stick to the current circular shape.
`holder.image.setBackgroundColor(Color.RED)` I think

PR #1435, opened 2020-09-11
Upgrade to Firebase Crashlytics

ahmedre:
The Crashlytics SDK is going to stop working on November 15th, 2020.
This patch upgrades to Firebase Crashlytics. It also removes most
analytics events for now (this will likely be re-added in the future in
sha' Allah). Fixes #1292.

PR #1458, opened 2020-10-17
Remove remaining crashlytics code

ahmedre:


ahmedre:
jazakAllah khair @ozbek - per your comment in #1435

PR #1463, opened 2020-10-17
Add an all results search suggestion

### LINKED ISSUE #1404: search improvements
searching for "سباتا" returns 1 suggestion - pressing enter gives 2 results (there should be 2 suggestions then, why only 1?)

ahmedre:
Improve search suggestions by adding an "all results" search suggestion
result. Because of the way Arabic search works, the search has to be
very broad, and more specific results have to be filtered in code.
Consequently, there's a limit of the first set of results found from the
database (before filtering). Consequently, queries that match many
things in the middle will only return some results instead of all. This
patch updates search suggestions to include an "all results" item to do
a full, unrestricted search.

Fixes #1404.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/data/QuranDataProvider.java
+++ b/app/src/main/java/com/quran/labs/androidquran/data/QuranDataProvider.java
@@ -133,6 +133,7 @@ public class QuranDataProvider extends ContentProvider {
 
     Context context = getContext();
     boolean gotResults = false;
+    boolean likelyHaveMoreResults = false;
     for (int i = start; i < total; i++) {
       if (gotResults) {
         continue;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/quran/labs/androidquran/data/QuranDataProvider.java
```diff
@@ -182,6 +182,12 @@ private Cursor getSuggestions(String query) {
       }
     }
 
+    if (context != null) {
+      mc.addRow(new Object[] {
```
ahmedre, line 186: maybe can improve this - only show for example if `queryIsArabic`, or if not all databases were searched, or if more than a certain number of results were found.

PR #1466, opened 2020-10-18
Use green color for qari with downloaded sura

benomaire:
Until we implement rearranging (issue 1448), I think this PR will be useful anyway (at least for me).

img


ahmedre (approved):

ahmedre:
جزاك الله خيراً

PR #1469, opened 2020-10-19
Use upright arrow for "go to" button in widget

### LINKED ISSUE #1467: Use upper right arrow for "go to" in widget
The widget is great! thanks to all who made it possible.
جزاكم الله خيرا

just a small note, I feel it will be more intuitive to use an arrow like the one in the picture below, because the current arrow looks like "back".
widget

@ismail-s @ahmedre @ozbek  What do you think?

benomaire:
Fixes #1467 

Should we remove `ic_baseline_arrow_forward_24.xml` ?


ahmedre:
جزاك الله خيراً
please do remove ic_baseline_arrow_forward_24.
also, now that i am thinking about it, these vectors will probably not work on older versions of Android - is it possible to see if the widget loads on the older versions? not sure if @ismail-s tested on older devices (i didn't unfortunately - only tested on more recent versions).

جزاك الله خيراً

ismail-s:
No, I didn't test on older versions.

ahmedre:
jazakumAllah khairan @benomaire 

in sha' Allah i'll test against older APIs and make sure it works before releasing.

ahmedre:
i tested on api 15 and api 17 emulators and it works - this was surprising to me, so i looked within the apk - turns out that aapt converted these to the respective pngs for all densities, which is why it works. so we're good here al7amdulillah.

PR #1471, opened 2020-10-24
Initial basic implementation of Analytics

ahmedre:
Implement analytics in such a way as to be able to swap out the
implementation with a NoopImplementation for cases where we don't want
to use Firebase.


ahmedre:
actual analytics events will be added in another patch in sha' Allah

PR #1486, opened 2020-11-06
Force blue colors for qari without downloaded sura

benomaire:
Fixes a bug with #1466. Sometimes other qari get the green color without downloaded sura.


ahmedre:
جزاكم الله خيراً

PR #1487, opened 2020-11-06
Update UI Strings

benomaire:
Mainly adding missing Arabic strings.

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -309,5 +309,41 @@
   <string name="more_translations">المزيد من الترجمات والتفاسير</string>
   <string name="menu_show_date">إظهار تاريخ الإضافة</string>
   <string name="pause">وقف مؤقت</string>
+  <string name="prefs_category_dual_screen">تفضيلات الصفحتين المتجاورتين</string>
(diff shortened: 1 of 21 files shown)


ahmedre:
جزاكم الله خيراً
let me know when this is ready can merge, barak Allah feekum

benomaire:
Please comment on adding (الوضع العربي) in English. If you don't like it, I will revert it. Otherwise, I think it should be added to other languages.

ahmedre:
sure, in the Arabic one, can we add the English in it as well (get a few people often asking "how do I get out of Arabic mode?" so that would help them also)

benomaire:
But I think that way is not unique! Should we just have English?

ahmedre:
yeah it's not ideal i guess - because depending on the language people have set, the string would be wrong. maybe we leave them as they are for now.

benomaire:
For Arabic,
How about using 
`إلغاء هذا الوضع يؤدي إلى استعمال لغة الجهاز الافتراضية`
instead of
`سيتم تحويل التطبيق إلى اللغة العربية`
?

ahmedre:
I like
سيتم تحويل التطبيق إلى اللغة العربية
more since it's clearer - that selecting this option will cause the app to be in Arabic.

benomaire:
If no other comments, feel free to merge.

ahmedre (approved):
جزاكم الله خيراً

PR #1500, opened 2020-11-26
Improve Header and Footer Text

### LINKED ISSUE #1322: Headers font size of sura name, juzaa and page number are small
Headers font size of sura name, juzaa and page number are small, as shown in the picture attached.
Screenshot_20200419-053445_Quran

benomaire:
Should fix #1322 

I am thinking of changing the Arabic font. Mainly so that `2` does not look like `3` (the Eastern Arabic Numbers). What do you think?
Also how about using `الجزء التاسع والعشرون` instead of `الجزء 29`?


benomaire:
Used [KFGQPC Naskh](https://fonts.qurancomplex.gov.sa/wp02/%d8%b9%d8%ab%d9%85%d8%a7%d9%86-%d8%b7%d9%87-%d9%86%d8%b3%d8%ae#05) font for Arabic.
I don't think I will change 29 to التاسع والعشرون. It will be long if we have ثلاثة أرباع الحزب.

ahmedre:
jazakAllah khairan - just tried this out and i like it!

two questions:
1. can you think of anyway to do this without the new font - i tried with our existing hafs font (uthmanic_hafs_ver12.otf) and the sura text looks the same - the only difference is the number of the juz' appears in an ayah marker. i think the font is not too big so it's ok to use, but if there's a way we can use the existing one, we'd avoid increasing apk size (or if you can think of other places in the future we'd want to use this new font, we can use both?).
2. i am a bit used to the old font color - why the change here? :) i am not against it, but just wondering what your thought was.

jazakumAllah khairan

ahmedre:
also, really sorry i forgot to look at this for the past few weeks and jazakumAllah khairan for doing this!

benomaire:
img1 (at the end) shows how "uthmanic_hafs_ver12.otf" looks on `Nexus 4 API 22`. The glyph of the comma is not an issue, since we can use a dash or just a space. However, the numbers are significantly smaller than the text!
Just changing the font to "UthmanTN1Ver10.otf", the numbers are the same size as text, and the comma appears naturally. See img2

I also tried using "kitab.ttf", and it looks like img3. The main problem I have with this is the shape of 2. If the license of the font permits changes, we can change it, and we get img4. Note the the glyph of `2` does not fit well, since I copied it from "UthmanTN1Ver10.otf".
One advantage for KFGQPC fonts is they are by Uthman Taha, and look similar to Quran Text. However, I think Kitab is better on Tafsir text, so we should keep it.

For the color, black is what is used in the printed Mushaf, and I have seen a suggestion to make it BOLD, I think this is because it was faint and small thus hard to see. However, since we don't have a border around the page, probably getting it Dark grey is better to distinguish it from Quran. I think slightly darker than what it used to be, for easy reading by all people. 

For the time being, I found an issue with dual-page mode where the header touches the quran text. I think reducing the font size and increasing the margin solve this.

I totally understand the delay, we all go through this :)

<details><summary>Images</summary>

img1:
img1

img2:
img2

img3:
img3

img4:
img4


</details>

ahmedre (approved):

ahmedre:
جزاكم الله خيراً
makes sense, let's go with this then - barak Allah feekum!

PR #1515, opened 2020-12-27
Add a fallback url for when main urls don't work

ahmedre:


app/src/main/java/com/quran/labs/androidquran/service/QuranDownloadService.java
```diff
@@ -464,10 +466,11 @@ private boolean downloadFileWrapper(String urlString, String destination,
       }
 
       String url = urlString;
-      if (i > 0) {
-        // let's try http instead of https?
-        if (urlString.contains("quran.com") || urlString.contains("quranicaudio.com")) {
-          url = urlString.replace("https://", "http://");
+      if (fallbackByDefault || i > 0) {
+        if (urlString.contains("android.quran.com")) {
```
ozbek, line 470: You could directly use `replace` without if block/check, because with `contains` and `replace` you are doing double look-up anyway.

app/src/main/java/com/quran/labs/androidquran/pageselect/PageSelectPresenter.kt
```diff
@@ -42,8 +43,10 @@ class PageSelectPresenter @Inject
         } else if (!downloadingSet.contains(it.key)){
           downloadingSet.add(it.key)
           val url = "$baseUrl/${it.key}.png"
+          val backupUrl = "$backupBaseUrl/${it.key}.png"
           compositeDisposable.add(
               imageUtil.downloadImage(url, previewImage)
+                  .onErrorResumeNext(imageUtil.downloadImage(backupUrl, previewImage))
```
ozbek, line 49: Elsewhere you are `replace`ing on error, here a separate variable. Maybe use a common convention?

Actually, I personally would prefer to have a single utility class/method for all `replace`ments :)

app/src/main/java/com/quran/labs/androidquran/service/QuranDownloadService.java
ahmedre: jazakAllah khairan - done

app/src/main/java/com/quran/labs/androidquran/pageselect/PageSelectPresenter.kt
ahmedre: jazakAllah khairan - done

PR #1517, opened 2020-12-29
Allow upgrade logic per page type

ahmedre:
Today, there's a lot of code within the initial data check and upgrade
path that's specific to a certain page type - for example, the database
is bundled for madani databases, the pages may be bundled in the future
for warsh databases, and so on.

This patch adds an interface wherein different page types can intercept
and handle updates on their own before going to the server for data.


common/data/src/main/java/com/quran/data/core/QuranFileManager.kt
```diff
@@ -0,0 +1,23 @@
+package com.quran.data.core
+
+import androidx.annotation.WorkerThread
+
+interface QuranFileManager {
+  @WorkerThread
```
ahmedre, line 6: one day, these will become suspending functions in sha' Allah...

app/src/main/java/com/quran/labs/androidquran/presenter/data/QuranDataPresenter.kt
```diff
@@ -294,24 +277,6 @@ class QuranDataPresenter @Inject internal constructor(
     return quranDataStatus
   }
 
-  @WorkerThread
-  private fun copyArabicDatabaseIfNecessary() {
```
ahmedre, line 298: moved into the madani implementation as an upgrade

app/src/main/java/com/quran/labs/androidquran/presenter/data/QuranDataPresenter.kt
```diff
@@ -98,20 +95,6 @@ class QuranDataPresenter @Inject internal constructor(
     }
   }
 
-  private fun copyLocalDataIfNecessary(status: QuranDataStatus): Single<QuranDataStatus> {
```
ahmedre, line 101: was not used today - but was planning on using for warsh, so moved this as a warsh upgrade and removed from here.

PR #1519, opened 2020-12-30
Use NumberPicker for Range and Verse Repeat Selection

### LINKED ISSUE #985: Making a custom entries in the number of ayah looping

benomaire:
This is a replacement of #1220 Thanks to @fnzainal for starting this.
Should fix #985 
Now we have 99 options for repeating. And yes, we lost the suffix 'time' or 'times'.
For Arabic, I used a string array to flip the visual appearance of the horizontal number picker.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
@@ -3,7 +3,6 @@ package com.quran.labs.androidquran.ui.fragment;
 
 import android.content.Context;
 import android.graphics.Typeface;
-import android.os.Build;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -14,10 +13,10 @@ import android.widget.Button;
 import android.widget.CheckBox;
 
 import com.quran.data.core.QuranInfo;
+import com.quran.data.model.SuraAyah;
 import com.quran.labs.androidquran.QuranApplication;
 import com.quran.labs.androidquran.R;
 import com.quran.labs.androidquran.dao.audio.AudioRequest;
-import com.quran.data.model.SuraAyah;
 import com.quran.labs.androidquran.ui.PagerActivity;
 import com.quran.labs.androidquran.ui.helpers.HighlightType;
 import com.quran.labs.androidquran.ui.util.TypefaceManager;
@@ -25,6 +24,9 @@ import com.quran.labs.androidquran.util.QuranSettings;
 import com.quran.labs.androidquran.util.QuranUtils;
 import com.quran.labs.androidquran.view.QuranSpinner;
 import com.shawnlin.numberpicker.NumberPicker;
+
+import java.text.NumberFormat;
+import java.util.Locale;
 
 import javax.inject.Inject;
 
@@ -45,8 +47,7 @@ public class AyahPlaybackFragment extends AyahActionFragment {
   private QuranSpinner endingAyahSpinner;
   private NumberPicker repeatVersePicker;
   private NumberPicker repeatRangePicker;
-  private boolean isArabicNames;
-  private int numOfOptions;
+  private final int numOfOptions = 99;
   private CheckBox restrictToRange;
   private ArrayAdapter<CharSequence> startAyahAdapter;
   private ArrayAdapter<CharSequence> endingAyahAdapter;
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
```diff
@@ -58,13 +64,45 @@ public View onCreateView(LayoutInflater inflater,
...
+    repeatRangePicker = (NumberPicker) view.findViewById(R.id.repeat_range_picker);
+    int defaultVerseRepeat = 3;
+    int defaultRangeRepeat = 7;
 
     final Context context = getActivity();
+
+    // Using Eastern Arabic Numerals
+    isArabicNames = QuranSettings.getInstance(context).isArabicNames();
+    if (isArabicNames){
+      String[] arNums = {"٩٩","٩٨","٩٧","٩٦","٩٥","٩٤","٩٣","٩٢","٩١","٩٠","٨٩","٨٨","٨٧","٨٦","٨٥","٨٤","٨٣","٨٢","٨١","٨٠","٧٩","٧٨","٧٧","٧٦","٧٥","٧٤","٧٣","٧٢","٧١","٧٠","٦٩","٦٨","٦٧","٦٦","٦٥","٦٤","٦٣","٦٢","٦١","٦٠","٥٩","٥٨","٥٧","٥٦","٥٥","٥٤","٥٣","٥٢","٥١","٥٠","٤٩","٤٨","٤٧","٤٦","٤٥","٤٤","٤٣","٤٢","٤١","٤٠","٣٩","٣٨","٣٧","٣٦","٣٥","٣٤","٣٣","٣٢","٣١","٣٠","٢٩","٢٨","٢٧","٢٦","٢٥","٢٤","٢٣","٢٢","٢١","٢٠","١٩","١٨","١٧","١٦","١٥","١٤","١٣","١٢","١١","١٠","٩","٨","٧","٦","٥","٤","٣","٢","١"};
```
ahmedre, line 80: instead of this, can we just do:

```
val numberFormatter = NumberFormat.getInstance(Locale.ARABIC);
numberFormatter.format(number)
```

instead of having this array of numbers? as a matter of fact, guessing if we don't pass the locale, it gets it from the default locale, which we should be overriding to Arabic already.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahPlaybackFragment.java
```diff
@@ -58,13 +64,45 @@ public View onCreateView(LayoutInflater inflater,
...
+      numOfOptions = arNums.length;
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
ahmedre, line 82: probably don't need a version check here since these are just numbers - only restricting the tafseer text because the tashkeel renders strange on older versions, numbers should be ok though

app/src/main/res/layout-ar-land-v17/audio_panel.xml
```diff
@@ -96,17 +96,30 @@
             android:text="@string/play_verses_range"
             android:textAppearance="@style/PanelLabel"
             />
-        <com.quran.labs.androidquran.view.QuranSpinner
-            android:id="@+id/repeat_range_spinner"
+        <com.shawnlin.numberpicker.NumberPicker
```
ahmedre, line 99: what do you think about using he default Android `NumberPicker`? the only downside is we can't style it on api levels before 24, but i think that is ok?

ahmedre (commented):
haven't tested this yet, but some initial comments

ahmedre:
also jazakumAllah khairan for pushing this forward!

benomaire:
Replying to all your very resonable comments:
1. I want a **horizontal** number picker, hopefully looking the same on all devices. [This answer](https://stackoverflow.com/a/6797248) looks discouraging for the default NumberPicker. The used NumberPicker is easily customizable and works on API 15.
2. I agree the string array is very ugly. But the main reason is I flip the numbers. So the NumberPicker looks to be from right to left (for Arabic users). I actually did not notice `setOrder`, and that is mainly why I did that ugly solution.

So the only thing remaining is whether we keep the external NumberPicker.

ahmedre (approved):
جزاكم الله خيراً
really sorry for the very late review - just tried this and it looks good, barak Allah feekum - one minor thing - may we make the font size smaller in English? please see:

Screenshot_1612044041

also - maybe let's make the max 25? 99 seems a bit excessive?
جزاكم الله خيراً

ahmedre:
i made a small change here - combined all these layouts into 2 layouts - one for `play_each_verse` and one for `play_set_of_verses` so we can change one instead of changing all of them

ahmedre:
merged at b8ebffe8d358f839c69e8a90a3a3958e827ac568 and some other commits.
please do see if you can fix the font size in English though - maybe a dimen for the text size that's different for Arabic and English - jazakumAllah khairan

benomaire:
> also - maybe let's make the max 25? 99 seems a bit excessive?

But you had the infinite `loop` before :)

PR #1520, opened 2020-12-30
Remove old upgrades

ahmedre:
There are some very old preference upgrades that still run. These are
actually very dangerous on non-Madani releases, since they may result in
the data being "lost" if the default data directory changes. Whenever
new updates like this are added, care should be taken to treat different
flavors differently.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
+++ b/app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
@@ -185,63 +185,6 @@ public class QuranSettings {
         version = prefs.getInt(Constants.PREF_VERSION, 0);
       }
 
-      if (version != 0) {
-        if (version < 2672) {
-          // migrate preferences
-          setAppCustomLocation(prefs.getString(Constants.PREF_APP_LOCATION, null));
-
-          if (prefs.contains(Constants.PREF_SHOULD_FETCH_PAGES)) {
-            setShouldFetchPages(prefs.getBoolean(Constants.PREF_SHOULD_FETCH_PAGES, false));
-          }
-
-          if (prefs.contains(QuranDownloadService.PREF_LAST_DOWNLOAD_ERROR)) {
-            setLastDownloadError(
-                prefs.getString(QuranDownloadService.PREF_LAST_DOWNLOAD_ITEM, null),
-                prefs.getInt(QuranDownloadService.PREF_LAST_DOWNLOAD_ERROR, 0));
-          }
-
-          prefs.edit()
-              .remove(Constants.PREF_VERSION)
-              .remove(Constants.PREF_APP_LOCATION)
-              .remove(Constants.PREF_SHOULD_FETCH_PAGES)
-              .remove(QuranDownloadService.PREF_LAST_DOWNLOAD_ERROR)
-              .remove(QuranDownloadService.PREF_LAST_DOWNLOAD_ITEM)
-              .remove(Constants.PREF_ACTIVE_TRANSLATION)
-              // these aren't migrated since they can be derived pretty easily
-              .remove("didPresentPermissionsRationale") // was renamed, removing old one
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -185,63 +185,6 @@ public void upgradePreferences() {
         version = prefs.getInt(Constants.PREF_VERSION, 0);
       }
 
-      if (version != 0) {
-        if (version < 2672) {
```
ahmedre, line 189: this was from 2016.

app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -185,63 +185,6 @@ public void upgradePreferences() {
...
-              .remove(QuranDownloadService.PREF_LAST_DOWNLOAD_ERROR)
-              .remove(QuranDownloadService.PREF_LAST_DOWNLOAD_ITEM)
-              .remove(Constants.PREF_ACTIVE_TRANSLATION)
-              // these aren't migrated since they can be derived pretty easily
-              .remove("didPresentPermissionsRationale") // was renamed, removing old one
-              .remove(Constants.PREF_DEFAULT_IMAGES_DIR)
-              .remove(Constants.PREF_HAVE_UPDATED_TRANSLATIONS)
-              .remove(Constants.PREF_LAST_UPDATED_TRANSLATIONS)
-              .apply();
-        } else if (version < 2674) {
```
ahmedre, line 216: so was this.

app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -185,63 +185,6 @@ public void upgradePreferences() {
...
-          // a bad state due to not crashing in 2.6.7-p2 (thus getting its incorrect behavior),
-          // and thus crashing on 2.6.7-p3 and above (where the bug was fixed). this works around
-          // this issue.
-          try {
-            getLastDownloadItemWithError();
-            getLastDownloadErrorCode();
-          } catch (Exception e) {
-            clearLastDownloadError();
-          }
-        } else if (version == 2800) {
```
ahmedre, line 227: 2017

app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -185,63 +185,6 @@ public void upgradePreferences() {
...
-          // upgrading from 2800, need to remove notification channels
-          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-            deleteOldNotificationChannels();
-          }
-        }
-
-        if (version < 2943) {
```
ahmedre, line 234: just cleanup, ok if we don't do it.

app/src/main/java/com/quran/labs/androidquran/util/QuranSettings.java
```diff
@@ -185,63 +185,6 @@ public void upgradePreferences() {
...
-          prefs.edit().remove("didDownloadPages").apply();
-        }
-
-        if (version < 2961) {
```
ahmedre, line 238: more recent but ok if we don't do it anymore since al7amdulillah most people stopped complaining about partial pages now.

ahmedre (commented):
the main reason for this diff though isn't cleanup - it's because naskh's versions (and Qaloon and Warsh's) are much lower than Quran for Androids, so every update of any of those flavors runs through all this upgrade logic.

this is especially dangerous in cases like this release, where the default directory is changed to internal storage - you have the data on one directory, but this upgrade "clears" the setting for it, so the app looks on internal storage (the new default), doesn't find the data, and asks you to redownload, even though you have the data.

PR #1527, opened 2021-01-06
Update Uzbek UI localization strings

ozbek:


app/src/main/res/values-uz/strings.xml
```diff
@@ -90,6 +92,8 @@
     </string>
   <string name="search_hint">Izlash</string>
   <string name="quranSearchType">Suralar</string>
+  <string name="search_full_results">Barcha natijalar</string>
```
ozbek, line 95: I could not find use case for this string (and one below). From the source logic, it looks like it is only used in the Arabic locale. Maybe put the Arabic translations in the default locale strings and make them non-translatable? What do you think @ahmedre?

app/src/main/res/values/strings.xml
```diff
@@ -224,8 +224,8 @@
   <string name="prefs_export_summary">Export a copy of bookmarks and tags</string>
   <string name="prefs_page_type_title">Page Type (experimental)</string>
   <string name="prefs_page_type_summary">Select the type of reading pages</string>
-  <string name="prefs_sura_translated_name_summary">Show sura translated name</string>
-  <string name="prefs_sura_translated_name_title">Sura translated name</string>
+  <string name="prefs_sura_translated_name_summary">Show the translation of surah name</string>
```
ozbek, line 227: Synced with the `surah` usage elsewhere (also minor wording change).

ozbek:
Travis build has failed at `com.quran.labs.androidquran.ui.QuranActivityTest > testClickingOnSuraOnListViewNavigatesToSura[test(AVD) - 5.1.1]` but the test is OK on my local machine :\

app/src/main/res/values-uz/strings.xml
ahmedre: yeah that sounds reasonable today - but am wondering if we should have something similar for non-Arabic cases where the number of results is a lot more than what we'd show in the suggestions instead.

ahmedre:
jazakumAllah khairan!

PR #1528, opened 2021-01-06
Disable split page prefs if dual page mode is not enabled

ozbek:
Because split page requires dual pages and does not work without it.


app/src/main/res/xml/quran_preferences.xml
```diff
@@ -48,7 +48,7 @@
         android:key="@string/prefs_sura_translated_name"
         android:persistent="true"
         android:summary="@string/prefs_sura_translated_name_summary"
-        android:title="@string/prefs_sura_translated_name_summary"
+        android:title="@string/prefs_sura_translated_name_title"
```
ozbek, line 51: This is a non-related change, but because it is a minor modification, included in this PR.

ahmedre (approved):
jazakumAllah khairan

Issue #1529, opened 2021-01-07
add ability to increase playback speed of quran 


MahmoudMabrok:
i did it, i will make PR ISA

benomaire:
I used to wish we have this feature. But if you think about it, missing this feature is probably the better choice. Some reasons are:
1. I don't think going over Quran **fast** is *an objective*. For example, refer to Tafseer of 7:204, 17:106, 25:32, 73:4, and 75:16.
2. I am afraid if people get used to the fast speed, they cannot listen to the true speed anymore.
3. I don't know of any reputable radio broadcast that aired fast-forwarded Quran audio (not even casual human speech AFAIK).
4. Reciters (Qaris) spend years on end learning Quran Tarteel and Tajweed, and recordings take at least a month (if recorded in Ramadhan for example), and studio recordings are extensively reviewed. All these efforts would be in vain if people speed up and loose the chance of learning correct Tajweed, feeling the meanings, etc.
5. If someone wants to memorize fast, then you know the companions of prophet Mohammad Peace be upon him used to go 10 verses at a time, memorizing and learning at the same time. Hopefully, if you pass away before finishing, your reward is saved because you were on the path. إن شاء الله
6. Different reciters with different *natural* speeds are available.
7. sorry to mention this analogy but I think people listening to music don't speed it up.

I know there are arguments about reading Quran in like 3 days. But I don't think this is the norm or recommendation. I also know people speed up using other means, but with the app we share responsibility as well, so I think better safe than sorry.
What do you think? 
Should we add more Taraweeh recitations (for haramain, that will not be a single qari)?
جزاكم الله خيرا

ahmedre:
well said @benomaire, jazakumAllah khairan - lots of people requested this over time, but never entertained it for many of the reasons you mentioned, and don't want to - i don't mind adding a "7adr" mus7af recitation where the reciter in general reads with a natural quick voice (or more tarweeh type recordings).

MahmoudMabrok:
Thanks for this discussion, I added it as i had a "جلقة تحفيظ"  and i was need to listen ayahs in speed so i can re-memorize my recitation, i will need this option only before go to shekh (الشيخ).

amynbe:
Not that I need this feature (I'm more of a visual learner, listening doesn't help my memory as much), but just to explore more options:
- A possibility would be to hide this feature behind a setting labelled "Advanced player interface".
- A warning could further discourage the user from enabling the option.
[...]

PR #1530, opened 2021-01-08
feat: add speedup & down to sound player UI

MahmoudMabrok:
implement the feature at #1529


princeparacha58:
Assalam mu alikum brother, i need speed up and speed down in quran andriod audio player please add in your software jazakallah

MahmoudMabrok:
@princeparacha58 I think now we have `hadr` recitation, so if still need this feature, we can have a build from this PR, but I think it will not have all data sources in current app version.

closed as  `hadr` recitation added.

ahmedre:
let's keep it open, thinking to merge it but with guidelines on what max speed to set depending on the Qari's rate of recitation, since this remains the top requested feature.

MahmoudMabrok:
if will be merged @ahmedre I would spend today and rest of week on working on it. 

I really want to contribute and help Quran users.

ahmedre:
planning to merge it in sha' Allah - جزاكم الله خيراً for building it and really sorry for being quiet about it all this time (as you know, I was against this idea from earlier on, but after discussions and brother @mohamede1945's idea to keep things beneficial, and it being the top feature request), planning on seeing this through in sha' Allah.

MahmoudMabrok:
No problem my brother, so I think we can make make 1.5(as max to all recitation) until we get a comprehensive solution to be related to Quran recitation type, do you agree @ahmedre ? 

I follow rule said `write, revisit, refactor, enhance`

ahmedre:
Sure, that sounds reasonable. Cc @mohamede1945 fyi

MahmoudMabrok:
@ahmedre  I have pushed the changes. please check

ahmedre (commented):
overall looks good, جزاكم الله خيراً

a few minor things:
1. we probably should have some pre-sets instead (ex 0.5x, 0.75x, 1x, 1.25x, 1.5x) instead of a dynamic increase/decrease.
2. if this only works on M+, let's not show the ui components for versions less than M

MahmoudMabrok:
@ahmedre  these pre sets will be in code only or will reflect in UI.

ahmedre:
we'd need to reflect in the UI also. for the audio bar, we can do the same as repeat (if you tap the repeat icon, the number cycles there between the valid repeat amounts). then, we can also expose a similar layout to repeat in the settings bottom sheet.

mohamede1945:
> Sure, that sounds reasonable. Cc @mohamede1945 fyi

Sounds great. Jazak Allah Khyran. And thanks brother @MahmoudMabrok for implementing it.

BTW, Here is the list that the quran.com website shows. I think this PR will be just for [0.5, 0.75, 1, 1.25, 1.5], right?

image

ahmedre:
جزاكم الله خيراً
Did some minor clean up - it's still not ready yet, need to figure out a fix for sura transitions where the speed delta gets lost when one sura ends and another starts to play. Likely need to put the speed inside the `AudioRequest` itself. Will look over the weekend or next week in sha' Allah.

MahmoudMabrok:
I tried to add to `AudioRequest ` but faced a lot of issues. 
I will check ISA.

ahmedre:
@MahmoudMabrok I've fixed some issues (including when we transition suras, the playback speed is now properly maintained). I still see two things before we can merge this:

1. the bigger problem - repeat seek timings are off when repeating at a higher play rate - ex play and press the button to repeat the ayah (and make speed 1.5x) - you'll notice that it reads the ayah and some of the next before going back (at 1x it's fine). note that "seek" itself is fine - i.e. going to the next or previous ayah goes to the correct time stamp.
2. nice to have - expose this in the audio panel settings - I've added the wiring, just needs the ui piece.

I think the first is caused by our audio check being at 1x - we need to run this more (or less) frequently to fix. the second one just needs to be done. just updating you so you know.

ahmedre:
aforementioned changes made al7amdulillah.
جزاكم الله خيراً @MahmoudMabrok

MahmoudMabrok:
Thanks so much for your efforts and merge.

PR #1534, opened 2021-01-08
Update README.md

MahmoudMabrok:
add code analysis
it list each language used in code/project files, count pure lines, blank lines


ahmedre (approved):

ahmedre:
didn't realize this was a static image - maybe no point really then? :(

MahmoudMabrok:
what issue of being static image, 
do u want it as dynamic feature
@ahmedre

ahmedre:
we already know the stats of percentage via Github side bar - lines of code is not important to be shared on a page like this imho

MahmoudMabrok:
i thought display LOC may be an indication of how big is our app

PR #1514, opened 2020-12-26
Add Bulk Download Dialog

benomaire:
Possible use cases:
1. Download from Surah 18 to Surah 114
2. Download from Surah 78 to Surah 114
3. etc

This saves many touches for highlighting all these Surahs.

PR #1540, opened 2021-01-31
Use large text size only for Arabic in NumberPicker

benomaire:
Used a smaller default text size, and increased it for Arabic.
Also implemented the comments in #1514


ahmedre (approved):

ahmedre:
جزاكم الله خيراً

PR #1544, opened 2021-02-11
Fix French translation and add French sura names

azeerman:


ahmedre:
jazakumAllah khairan!

PR #1546, opened 2021-02-11
Update dependencies

ahmedre:


ahmedre:
This patch also removes Jetifier.

PR #1565, opened 2021-03-20
Update versions of dependencies

ahmedre:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-bin.zip
```
ozbek, line 3: FWIW, Android Studio favors `-all` package than `-bin`.

ahmedre: true, but been reading in some circles that there's really no point because `-all` gets sources, whereas `-bin` gets sources only upon demand - links for reference:

https://twitter.com/CedricChampeau/status/1113871652809531392
https://twitter.com/CedricChampeau/status/950376868716470273
https://youtrack.jetbrains.com/issue/IDEA-203597

ozbek: TIL.
Hope the `-all` suggestion notification is going away sooner than later :)

Issue #595, opened 2015-12-10
Replace "infinitely" with "loop", "indefinitely" or "continuously"


ozbek:
Although I can not say that's incorrect, option name "infinitely" (the last item in `repeatValues`) is not a common word used for playback repetition. Normally, the words "loop", "indefinitely" or "continuously" are used in regards to playback.

What do you think?

ahmedre:
sure - i like both "loop" and "indefinitely," but i think "loop" is probably better.

ahmedre:
jazakAllah khairan for fixing this.

PR #1583, opened 2021-04-09
add a checkbox for infinite repetition

### LINKED ISSUE #1582: Add back infinite repeat setting
people are complaining that the infinite setting is removed.
cc @benomaire if you want to pick this one up

benomaire:
should fix #1582
Feel free to implement any other option, or improve on this one.

### DIFF
--- a/app/src/main/res/layout/play_each_verse.xml
+++ b/app/src/main/res/layout/play_each_verse.xml
@@ -15,7 +15,7 @@
       />
   <com.shawnlin.numberpicker.NumberPicker
       android:id="@+id/repeat_verse_picker"
-      android:layout_width="wrap_content"
+      android:layout_width="110dp"
       android:layout_height="40dp"
       app:np_orientation="horizontal"
       app:np_width="135dp"
@@ -31,4 +31,11 @@
       app:np_fadingEdgeEnabled="true"
       app:np_wrapSelectorWheel="true"
       />
+
+  <CheckBox
+      android:id="@+id/checkBoxInfiniteVerse"
+      android:layout_width="wrap_content"
+      android:layout_height="match_parent"
+      android:layout_weight="1"
+      android:text="@string/infinity" />
 </LinearLayout>
--- a/app/src/main/res/layout/play_each_verse.xml
+++ b/app/src/main/res/layout/play_each_verse.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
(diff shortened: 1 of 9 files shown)


app/src/main/res/layout/play_each_verse.xml
```diff
@@ -31,4 +31,11 @@
       app:np_fadingEdgeEnabled="true"
       app:np_wrapSelectorWheel="true"
       />
+
+  <CheckBox
+      android:id="@+id/checkBoxInfiniteVerse"
+      android:layout_width="wrap_content"
+      android:layout_height="match_parent"
+      android:layout_weight="1"
+      android:text="@string/infinity" />
```
ozbek, line 40: Although `infinity` is an attractive word, it sounds weird in some language translations. Also, we had this issue #595.
I suggest to change it back to `loop`.

app/src/main/res/layout/play_set_of_verses.xml
```diff
@@ -33,4 +33,11 @@
     app:np_fadingEdgeEnabled="true"
     app:np_wrapSelectorWheel="true"
     />
+
+  <CheckBox
+      android:id="@+id/checkBoxInfiniteRange"
```
ozbek, line 38: Looks like you are using 4-space tabbing, Quran app uses 2-space tabbing.

app/src/main/res/layout/play_each_verse.xml
benomaire: `@string/infinity` is not translatable and [is just the Unicode character](https://github.com/quran/quran_android/blob/eb557e823bd2eb48f064f19108d906aa787d4955/app/src/main/res/values/strings.xml#L290) (∞). I don't mind using actual words (loop, etc), but don't know if all languages will fit, since the checkbox is next to the horizontal number picker.

app/src/main/res/layout/play_set_of_verses.xml
benomaire: Although i was using 4-space tabbing, this is probably the [CONTINUATION_INDENT_SIZE](https://github.com/quran/quran_android/blob/eb557e823bd2eb48f064f19108d906aa787d4955/quran_android-code_style.xml#L350), and reformatting the file increases all other lines above like this one :)

ahmedre:
جزاكم الله خيراً
some minor comments - and in sha' Allah whomever of us gets to this first can apply the changes in sha' Allah:

* this doesn't interplay with the existing setting - what i mean is if you tap the screen and tap the repeat icon 3 times so it shows infinite, that doesn't cause the x to be checked for infinite.

* one other thing i think we broke was automatically checking the "only play the above verses" box whenever "set of verses" changes to anything other than 1.

* also, maybe checking the infinite box should gray out the control?

جزاكم الله خيراً

app/src/main/res/layout/play_each_verse.xml
ozbek: I see, thanks. I have not noticed `infinity` was a symbol.

ahmedre:
working on this now.

ahmedre:
جزاكم الله خيراً
merged this and updated it - ended up using strings 😂

ahmedre:
i know you had it as strings initially - i was wrong 😂
جزاكم الله خيراً @benomaire

PR #1601, opened 2021-05-05
Add jump to shortcut

ismail-s:
- Add [a shortcut](https://developer.android.com/guide/topics/ui/shortcuts/creating-shortcuts#static) for launching the `ShowJumpFragmentActivity`
- Add espresso tests for `ShortcutActivity`
- Do a minor improvement to the existing espresso test
- Kotlinize `ShortcutsActivity` 

Note that to test this, I found I had to add a debug suffix (see d3d334d), otherwise the shortcut would launch the activity from the app installed from the Play Store.


ahmedre:
jazakAllah khairan for doing this - lgtm

ahmedre (approved):
sorry it took so long and jazakAllah khairan

ahmedre:
> Note that to test this, I found I had to add a debug suffix (see d3d334d), otherwise the shortcut would launch the activity from the app installed from the Play Store.

will fix this in sha' Allah since it also happens for the existing one.

Issue #1620, opened 2021-07-03
Make sure translations in inline view are sorted properly


ahmedre:
Make sure the translations in the inline translation view (the one where you long press an ayah and view it) are sorted according to the sort order of the translations. presumably this isn't the case today.

MohAbbasher:
@ahmedre please close this issue

ahmedre:
جزاكم الله خيراً

PR #1626, opened 2021-07-17
Fix#1620 translations inline view ordered properly

MohAbbasher:


app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -185,9 +185,9 @@
   private ViewGroup.MarginLayoutParams audioBarParams;
   private boolean isInMultiWindowMode;
 
-  private String[] translationItems;
-  private List<LocalTranslation> translations;
-  private Set<String> activeTranslations;
+  private String[] translationSpinnerTitles;
```
ahmedre, line 188: how about `translationNames` instead?

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -185,9 +185,9 @@
...
+  private List<LocalTranslation> allTranslationsFromDb;
```
ahmedre, line 189: maybe `translations` is better since if we decide to read it from somewhere else (preferences, etc, etc) this variable won't be correctly named.

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -1123,16 +1123,16 @@ public void startTranslationManager() {
         }
       };
 
-  public List<LocalTranslation> getTranslations() {
-    return translations;
+  public List<LocalTranslation> getAllTranslationsFromDb() {
```
ahmedre, line 1126: and let's rename these accordingly

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
```diff
@@ -37,7 +38,7 @@
   private View translationControls;
   private QuranSpinner translator;
   private TranslationsSpinnerAdapter translationAdapter;
-  private List<LocalTranslation> translations;
+  private List<LocalTranslation> allTranslationsFromDb;
```
ahmedre, line 41: let's keep it `translations`?

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
```diff
@@ -111,34 +112,35 @@ public void refreshView() {
...
+            pagerActivity.getTranslationSpinnerTitles(),
+            allTranslationsFromDb,
+            activeTranslationsFilesNames,
             selectedItems -> {
               quranSettings.setActiveTranslations(selectedItems);
               refreshView();
             });
         translator.setAdapter(translationAdapter);
       }
+      else{
```
ahmedre, line 141: nit - space after else

app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.java
```diff
@@ -31,12 +31,12 @@
...
   }
 
-  public void refresh(VerseRange verseRange) {
+  public void refresh(List<String> orderedTranslationsFilesNames, VerseRange verseRange) { //  pass orderedTranslationsFilesNames here to return ordered Verses from DBs
     if (getDisposable() != null) {
       getDisposable().dispose();
     }
 
-    setDisposable(getVerses(false, getTranslations(quranSettings), verseRange)
+    setDisposable(getVerses(false, orderedTranslationsFilesNames, verseRange)
```
ahmedre, line 39: maybe instead of this (we can revert this change back), we can update `getVerses`, since it's used by both the normal and inline translation views - let's update that `getVerses` to do the sorting instead at the end via:

```java
final LocalTranslation[] sortedTranslations = Arrays.copyOf(translations, translations.length);
Arrays.sort(sortedTranslations, new LocalTranslationDisplaySort());
```

just as the code in `TranslationView` is doing today - we can then remove that code from `TranslationView` and have it in one place.

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
```diff
@@ -111,34 +112,35 @@ public void refreshView() {
...
+        translationAdapter.updateItems(pagerActivity.getTranslationSpinnerTitles(), allTranslationsFromDb, activeTranslationsFilesNames );
```
ahmedre, line 142: do we actually need this else though?

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
```diff
@@ -149,7 +151,18 @@ public void refreshView() {
...
+
+      for (LocalTranslation translation : allTranslationsFromDb) {
+        for (int i = 0; i < translationsFileNames.size(); i++) {
+          if (translation.getFilename().equals(translationsFileNames.get(i))) {
+            orderedTranslationsFilesNames.add(translationsFileNames.get(i));
+          }
+        }
+      }
+      translationPresenter.refresh(new ArrayList<>(orderedTranslationsFilesNames), verseRange);
     }
```
ahmedre, line 166: let's remove this as per my comment above

app/src/main/java/com/quran/labs/androidquran/view/InlineTranslationView.java
```diff
@@ -20,6 +20,7 @@
 import com.quran.labs.androidquran.common.TranslationMetadata;
 import com.quran.labs.androidquran.util.QuranSettings;
 
+import java.util.Arrays;
```
ahmedre, line 23: unused?

ahmedre (commented):
jazakumAllah khairan for this PR - some comments in line

ahmedre:
if you rebase, ui tests should also work now

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
MohAbbasher: it was hard to me to understand the code since there is 3 or 4 variables related to `translations` so I renamed this variables , but it's OK to revert it back. 

I think it will better if we add `isActive` column to `translations' table, it will be more concise and readable.

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
MohAbbasher: ok

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
MohAbbasher: yes we need it, if translations list order changed from `translationManagerActivity` the spinner change a accordingly

app/src/main/java/com/quran/labs/androidquran/ui/fragment/AyahTranslationFragment.java
MohAbbasher: ok

app/src/main/java/com/quran/labs/androidquran/view/InlineTranslationView.java
MohAbbasher: ok will remove it .
 ان شاء الله

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
MohAbbasher: we can also start migrating gradually  to `Room` ,`Hilt` , `livedata`, `MVVM` , is there in plan for that , where can we start ?

app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.java
MohAbbasher: yes it's better

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
ahmedre: This project started over 10 years ago, which is why there's so much legacy code here - in sha' Allah planning to modernize, just need to allocate some more time to it.

* Hilt - we're already using Dagger, and I prefer Dagger + Anvil from Hilt from usage in other projects. 

* Room - either it or SQLDelight, there's an issue somewhere on the tracker about it. Not sure which yet.

* LiveData - prefer not to use it - we should be moving more towards unidirectional flows though, which most of the new code is doing.

* MVVM - doesn't really matter VM vs P as long as we are respecting things (unidirectional flows, streams for communications, etc) and keeping the business logic layer testable.

MohAbbasher: I think we can start with converting all code to Kotlin

MohAbbasher: how about coroutine ?

app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.java
MohAbbasher: if we add `isActive` column to translations table, it will be better , what do you think ?

MohAbbasher: or we can reuse `activeTranslationsFilesNames` from
https://github.com/MohAbbasher/quran_android/blob/7819c5cf5e9f8f5299bf91b1eb45c229ffe5e29c/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java#L1405

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
ahmedre: re: converting code to Kotlin - at 28.6% Kotlin now, just converting things when i am going to work in those classes, so definitely an on-going process. @wiryadev has been opening some PRs also recently.

re: coroutines - the more recently written code from this year is already using coroutines - plan is to deprecate Rx moving forward.

app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.java
ahmedre: adding a column makes sense especially now that the sorting info is in the database also - but in order to keep this patch small, reuse what's there for now and we can keep this for a new PR?

app/src/main/java/com/quran/labs/androidquran/presenter/translation/BaseTranslationPresenter.kt
```diff
@@ -34,11 +34,24 @@ internal open class BaseTranslationPresenter<T> internal constructor(
...
+    val sortedTranslations: List<LocalTranslation> = ArrayList(translations)
+    Collections.sort(sortedTranslations, LocalTranslationDisplaySort())
+
+    for (t in sortedTranslations) {
+      for (i in translationsFileNames.indices) {
+        if (t.filename == translationsFileNames[i]) {
+          orderedTranslationsFilesNames.add(translationsFileNames[i])
+        }
+      }
+    }
```
ahmedre, line 52: can we simplify this? something like:

```kotlin
val orderedTranslationsFileNames = sortedTranslations
   .filter { translationFileNames.contains(translation.filename) }
   .map { it.filename }
```

ahmedre (approved):

ahmedre:
JazakumAllah khairan

app/src/main/java/com/quran/labs/androidquran/presenter/translation/BaseTranslationPresenter.kt
MohAbbasher: yes its better , will do that ان شاء الله

MohAbbasher:
fix #1620

PR #1627, opened 2021-07-17
Downgrade emulator to api 27

ahmedre:


ahmedre:
ui tests are hanging right now due to ReactiveCircus/android-emulator-runner#168. using suggested workaround to downgrade to api 27.

PR #1628, opened 2021-07-18
Migrate `NetworkModule` to Kotlin

wiryadev:


ahmedre:
jazakumAllah khairan!

PR #1633, opened 2021-07-19
migrate `ArabicDatabaseUtilsTest` to kotlin

wiryadev:
also i remove the third `assertThat` because it is the same with first one (see attached image)
image


ahmedre:
i think that probably should be
`assertThat(result.get(2).getAyahText()).isNotEmpty()`

app/src/test/java/com/quran/labs/androidquran/model/translation/ArabicDatabaseUtilsTest.kt
```diff
@@ -0,0 +1,123 @@
...
+      override fun getArabicDatabaseHandler(): DatabaseHandler {
+        return arabicHandler
+      }
+
+      override fun getAyahTextForAyat(ayat: List<Int>): Map<Int, String> {
+        val result: MutableMap<Int, String> = HashMap()
+        for (ayahId in ayat) {
+          result[ayahId] = "verse $ayahId"
+        }
+        return result
```
ahmedre, line 80: how about

```kotlin
return ayat.map { it to "verse $it" }.toMap()
```

app/src/test/java/com/quran/labs/androidquran/model/translation/ArabicDatabaseUtilsTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    val result = StringBuilder()
+
+    for (i in 0 until words) {
+      if (i > 0) {
+        result.append(" ")
+      }
+      result.append("word").append(i)
+    }
+
+    return result.toString()
```
ahmedre, line 108: how about:

```kotlin
return words.map { "word$it" }.joinToString(" ")
```

ahmedre:
jazakumAllah khairan - super minor comments

wiryadev:
> i think that probably should be
> `assertThat(result.get(2).getAyahText()).isNotEmpty()`

added in latest commit brothers

app/src/test/java/com/quran/labs/androidquran/model/translation/ArabicDatabaseUtilsTest.kt
wiryadev: words is integer so it can not be mapped like that i think?

ahmedre: sorry i meant `(0 until 5).map { ...`

wiryadev: i see, commit has been pushed, thanks for suggesting that

app/src/test/java/com/quran/labs/androidquran/model/translation/ArabicDatabaseUtilsTest.kt
ahmedre: may you also do this one @wiryadev, then ready to merge - jazakumAllah khairan!

wiryadev: sure i can, but i still use `ayahId`  instead of `it` for readability purpose, so it would be like this. Is it okay?
```
return ayat.map { ayahId -> ayahId to "verse $ayahId" }.toMap()
```

ahmedre:
jazakumAllah khairan!

PR #1635, opened 2021-07-21
Migrate some presenter file to Kotlin

wiryadev:


wiryadev:
this commit already included [here](https://github.com/quran/quran_android/pull/1636)

PR #1651, opened 2021-08-02
Unnecessary SDK check clean up

wiryadev:
*it may be looks big but it only contains SDK check removal


ahmedre (approved):
JazakumAllah khairan

PR #1652, opened 2021-08-02
more minor improvement based on code inspection

wiryadev:
*looks big, but contains very small change in each commit


ahmedre:
this is great, jazakumAllah khairan!

PR #1653, opened 2021-08-02
migrate `Constants` to Kotlin

wiryadev:


wiryadev:
instrument test failed to get dagger dependency 

image

ahmedre:
jazakAllah khairan - not sure why you're getting those issues, but it compiled fine on CI and no artifacts have changed so this should be ok in sha' Allah.

PR #1665, opened 2021-08-09
replace `IntArray` with `Pair` for quarterArray

wiryadev:
I was looking into `MadaniDataSource` and found that the array is only contains 2 items. So i thinks it is better to use Pair instead for 2 reasons:
1. Less error probability
With integer array, there is probability to access index that never exists, which with pair this problem should not exist because it locks to only 2 item. Other case is we could access the wrong index since number is not idiomatic enough compared to `first` and `second` keyword. 

2. Consistency
In Presenter, we use `first` and `last` to access the array whereas in `AudioUtils` we use index to access it. With pair, we will need to access it with `first` and `second` since there is no index accessor like in array. It will create consistency accross file using it.

*this only applies if the items is only 2 items


ahmedre:
How about using `SuraAyah` instead of `Pair` here, since ultimately, this is just the ayah number where each quarter starts?

wiryadev:
i see, going to try that

wiryadev:
New PR is opened [here](https://github.com/quran/quran_android/pull/1667), close this now

PR #1671, opened 2021-08-11
optimize custom view class

wiryadev:
- make unmodified variable as final
- remove unnecessary SDK check
- remove unnecessary `TargetApi`
- simplify `if` statement
- replace deprecated method

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
+++ b/app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
@@ -11,7 +11,7 @@ import android.widget.RelativeLayout;
 import com.quran.labs.androidquran.R;
 
 public class FitSystemRelativeLayout extends RelativeLayout {
-  private static final boolean IS_PRE_KITKAT = Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT;
+  private static final boolean IS_PRE_KITKAT = false;
 
   private int lastTopInset;
   private View toolBarParent;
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
```diff
@@ -11,7 +11,7 @@
 import com.quran.labs.androidquran.R;
 
 public class FitSystemRelativeLayout extends RelativeLayout {
-  private static final boolean IS_PRE_KITKAT = Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT;
+  private static final boolean IS_PRE_KITKAT = false;
```
ahmedre, line 14: is it still used? if so can we remove it since that's a dead code path now?

ahmedre (approved):

ahmedre:
jazakumAllah khairan

app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
wiryadev: it is unused, even the overriden method `fitSystemWindows` is already deprecated since Kitkat. Maybe we should delete that method too?

ahmedre: we can't remove it just yet unfortunately even if it's deprecated until we migrate to the new ways of doing things - otherwise it'll break full screen mode in the app

wiryadev: Hi, i try to comment this out (not deleting whole `fitSystemWindows`) and it is okay/not breaking (in Pixel API 21)
```
//    if (IS_PRE_KITKAT && lastTopInset != insets.top) {
//      toolBarParent.requestLayout();
//      lastTopInset = insets.top;
//    }
```
And yes, removing the whole fitSystemWindows make weird fullscreen

ahmedre: ok maybe let's leave it like it was until i fix this properly to upgrade to the new ways of doing things and can remove this code?

wiryadev: yes sure

PR #1672, opened 2021-08-11
minor change in `AudioFocusHelper` and `BookmarkPresenterTest`

wiryadev:
- change variable name in `AudioFocusHelper` based on review
- minor readability improvement for `BookmarkPresenterTest`

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
+++ b/app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
@@ -1,12 +1,21 @@
 package com.quran.labs.androidquran.view;
 
 import android.content.Context;
+import android.graphics.Rect;
 import android.util.AttributeSet;
+import android.view.View;
 import android.widget.RelativeLayout;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/quran/labs/androidquran/view/FitSystemRelativeLayout.java
```diff
@@ -32,41 +22,4 @@ public FitSystemRelativeLayout(@NonNull Context context, AttributeSet attrs, int
     // needed to fix positioning of the ayah toolbar
     setLayoutDirection(LAYOUT_DIRECTION_LTR);
   }
-
-  @Override
-  protected boolean fitSystemWindows(@NonNull Rect insets) {
```
ahmedre, line 37: i sadly don't think we can remove this yet (even if it's deprecated) because it'll break things (especially moving to/from full screen, etc). i really need to move the code to the more modern ways of doing things here, but please bring this back in the meanwhile.

ahmedre:
jazakumAllah khairan

PR #1673, opened 2021-08-12
rewrite `ServiceIntentHelper` in Kotlin and some minor change

wiryadev:
minor change:
- replace deprecated method
- remove some redundant initializer and make it `final`
- replace wrapper with primitive type
- extract common part from `if-else`


app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
```diff
@@ -445,7 +445,7 @@ private void handleIntent(Intent intent) {
   private class ReadGaplessDataTask extends AsyncTask<Integer, Void, SparseIntArray> {
 
     private int mSura = 0;
-    private String mDatabasePath = null;
+    private final String mDatabasePath;
```
ahmedre, line 448: can we rename to databasePath while you are here?

ahmedre (approved):

app/src/main/java/com/quran/labs/androidquran/service/AudioService.java
wiryadev: yes we can, changed opened [here](https://github.com/quran/quran_android/pull/1674)

Issue #1677, opened 2021-08-13
Bug: Add to bookmark doesnt work


wiryadev:
![image](https://user-images.githubusercontent.com/26060382/129345961-b5a81574-98fd-4923-a1ae-cb8561652863.png)

Bookmarks page button doesnt work. Tested in API 21, 29, and 30,
Looks like it from this [PR](https://github.com/quran/quran_android/pull/1644).
The logcat can be seen here: [Pastebin](https://pastebin.com/CS9mE48d).

**Update 1**
The log happens not when clicking the button, but when opening page. When the button is clicked, i dont see any log.

**Update 2**
The log is not valid since it still exist when there is no bug
![image](https://user-images.githubusercontent.com/26060382/129352571-c752299c-6de8-4e1a-a22e-f22b0ec92712.png)

**Update 3**
Only adding bookmark doesnt work, removing works

wiryadev:
**Update 4**
I have fixed this issue but can not push it yet due to it is late night now. I will open the PR tomorrow (in my timezone)

PR #1680, opened 2021-08-14
fix add bookmark bug and some refactor

wiryadev:
- fix [this](https://github.com/quran/quran_android/issues/1677) issue
- refactor writing style
- change some value type to be null-safety


ahmedre:
jazakumAllah khairan

PR #1682, opened 2021-08-14
remove SDK check and hacky method that used for Kitkat and below in `StorageUtils`

wiryadev:
removed due to `minSdk` now is Lollipop


ahmedre (approved):

ahmedre:
jazakumAllah khairan

PR #1688, opened 2021-08-19
Fix sura list subtitles

ahmedre:


ahmedre:
cc @wiryadev

wiryadev:
Thank you very much for the fix

PR #1691, opened 2021-08-20
rewrite `ShareUtil` in Kotlin

wiryadev:
share url, share ayah, and copy should works (i've tried it)


app/src/main/java/com/quran/labs/androidquran/util/ShareUtil.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    sb.append('-')
+      .append(quranDisplayData.getSuraAyahString(context, ayahInfo.sura, ayahInfo.ayah))
+    return sb.toString()
+  }
+
+  private fun getShareText(activity: Activity, verses: List<QuranText>): String {
+    val size = verses.size
+    val sb = StringBuilder("(")
```
ahmedre, line 84: there's a nicer way to do this in kotlin - https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/build-string.html

```
buildString() {
  append(...)
  // etc
}
```

ahmedre (approved):
jazakumAllah khairan

app/src/main/java/com/quran/labs/androidquran/util/ShareUtil.kt
wiryadev: i see, will open PR soon

PR #1713, opened 2021-09-24
Change compileSdk to 31 and targetSdk to 30

ahmedre:


ahmedre:
will need to stop /sdcard type directories from being accessible on Android 11 and above before releasing

PR #1717, opened 2021-09-25
Support new page types and fix some bugs

ahmedre:
This patch adds a field to PageProvider to handle different types of
pages - for example, in the future, pages could be rendered per line, or
using fonts, etc. It also increases the maximum number of files allowed
in a zip file limit.

This also fixes a bug that causes the page selection to crash after
scrolling to the very end.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/pageselect/PageSelectAdapter.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/pageselect/PageSelectAdapter.kt
@@ -92,7 +92,6 @@ class PageSelectAdapter(val inflater: LayoutInflater,
   }
 
   override fun destroyItem(container: ViewGroup, position: Int, o: Any) {
-    super.destroyItem(container, position, o)
     container.removeView(o as View)
   }
 }
--- a/app/src/main/java/com/quran/labs/androidquran/presenter/data/QuranDataPresenter.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/presenter/data/QuranDataPresenter.kt
@@ -203,7 +203,7 @@ class QuranDataPresenter @Inject internal constructor(
       }
 
       val pageType = quranSettings.pageType
-      if (!quranSettings.didCheckPartialImages(pageType)) {
+      if (!quranSettings.didCheckPartialImages(pageType) && !pageType.endsWith("lines")) {
         Timber.d("enqueuing work for $pageType...")
 
         // setup check pages task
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/quran/labs/androidquran/pageselect/PageSelectAdapter.kt
```diff
@@ -92,7 +92,6 @@ class PageSelectAdapter(val inflater: LayoutInflater,
   }
 
   override fun destroyItem(container: ViewGroup, position: Int, o: Any) {
-    super.destroyItem(container, position, o)
```
ahmedre, line 95: we shouldn't call super - it just throws an exception

app/src/main/java/com/quran/labs/androidquran/presenter/data/QuranDataPresenter.kt
```diff
@@ -203,7 +203,7 @@ class QuranDataPresenter @Inject internal constructor(
       }
 
       val pageType = quranSettings.pageType
-      if (!quranSettings.didCheckPartialImages(pageType)) {
+      if (!quranSettings.didCheckPartialImages(pageType) && !pageType.endsWith("lines")) {
```
ahmedre, line 206: moving forward, i don't want to support this partial pages concept nor downloading pages on the fly - so for new page types (lines, etc), just don't run the partial page checker. in the future, we should do this for all page types.

app/src/main/java/com/quran/labs/androidquran/util/QuranFileUtils.kt
```diff
@@ -145,9 +146,8 @@ class QuranFileUtils @Inject constructor(
         if (fileList == null) {
           Timber.d("haveAllImages: null fileList, checking page by page...")
           for (i in 1..totalPages) {
-            if (!File(dir, getPageFileName(i))
-                    .exists()
-            ) {
+            val name = if (oneFilePerPage) getPageFileName(i) else i.toString()
+            if (!File(dir, name).exists()) {
```
ahmedre, line 150: we're not checking **inside** the directories, but i think that's fine for now.

common/data/src/main/java/com/quran/data/source/PageContentType.kt
```diff
@@ -0,0 +1,5 @@
+package com.quran.data.source
+
+enum class PageContentType {
+  IMAGE, LINE
```
ahmedre, line 4: we can add other types in the future - ex `FONT` for font based rendering, etc.

PR #1718, opened 2021-09-25
localize title and translator

benomaire:


ahmedre:
جزاكم الله خيراً
the backend response today has a localized title that comes in, but that is usually in the language that the actual text is in - i agree with you that this needs to be done, but my initial thought is that i would prefer to do this on the backend now (update the payload to have the translations in whatever language is passed in if we know that language). what do you think?

PR #1776, opened 2021-11-04
Update sura names

### LINKED ISSUE #1739: Change Surahs name in Surah list and in page header
I hope we change the Surah names format in Surahs list and page header specially Arabic language (with Tashkil as real Mushaf)to new format as the table attached : [Surahs_names](https://drive.google.com/file/d/12hMJfqhlbd6o9Ekmi8co9-lPwwmBNlOb/view?usp=drivesdk)  

سُورَةُ الفَاتِحَةِ Al-Fâtihah AL-FÃTIHAH
سُورَةُ آلِ عِمۡرَانَ Âl-`Imrân ÃL-`IMRÃN

وجزاكم الله خيرا

ahmedre:
Fixes #1739.


app/src/main/res/values-fr/sura_names.xml
```diff
@@ -1,5 +1,122 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+  <string-array name="sura_names">
+    <item>AL-FÃTIHAH</item>
```
ahmedre, line 4: @nacer80 are these intentionally caps for people in France?

app/src/main/res/values-fr/sura_names.xml
```diff
@@ -1,5 +1,122 @@
...
+    <item>AL-ISRÃ\'</item>
+    <item>AL-KAHF</item>
+    <item>MARЇAM</item>
+    <item>TÃ-HÃ</item>
+    <item>AL-ANBIYÃ\'</item>
+    <item>AL-HAJJ</item>
+    <item>AL-MU\'MINOUN</item>
+    <item>AN-NOUR</item>
+    <item>AL-FURQÃN</item>
+    <item>ACH-CHU`ARÃ\'</item>
```
ahmedre, line 29: @nacer80 also, do all people who speak French use the CH instead of sh?

ahmedre:
not sure if arabic is looking too crowded now though
arabic

ahmedre:
cc @benomaire for review

nacer80:
السلام عليكم ورحمة الله 
How about "go to page" setting.
When we need to go to Surah, we have to write it with tashkil?

ahmedre:
وعليكم السلام ورحمة الله وبركاته
good point - should check that to make sure

nacer80:
Before we had an issue with "Hamza"
such as Surah "الإنسان" we have write with Hamza

app/src/main/res/values-fr/sura_names.xml
nacer80: Yes, in Franch language they use "ch" it pronanced as "sh" in English

app/src/main/res/values-fr/sura_names.xml
nacer80: We can use also lowercase, but the Mushaf printed by KFGQPC they used uppercase.


Also here they use uppercase 
https://quranenc.com/ar/browse/french_montada

ahmedre:
this still has a hamza problem that I have to fix

ahmedre:
fixed the hamza problem - but if you search with a hamza (i.e. if you type الإنسان), it won't find it. I guess I can / should fix that, but guessing this will be more rare

ahmedre:
fixed that also - the downside now is searching for الإ will return sura names where the alif is not with a kasra because we're just ignoring them - but I think this is ok.

with_cleanup

app/src/main/res/values-fr/sura_names.xml
benomaire: Related: Surah names used by NOOR International: https://www.noorinternational.net/fr/translation/swarlist
i did not check if they match the proposed names here

ahmedre: Noor International's is a bit different than the ones from @nacer80 - but I'll go ahead and merge. please let me know if there are any I should change afterwards.

PR #1840, opened 2022-01-10
update english sura_names_translated

benomaire:
The new list matches [Hilali&Khan 1441 print](https://www.almeshkat.net/book/15401). Except Surah 35, where I removed ", or The Angels" since we only have "Fâtir" instead of "Sūrat Fāṭir or Al-Malā’ikah". Other exceptions with similar arguments: Surah 40 and Surah 47 and Surah 76.

For Surah 47, I used "صَلَّى ٱللَّهُ عَلَيۡهِ وَسَلَّمَ" instead of "".

I also changed "Al-Mujâdalah" to "Al-Mujâdilah" to match "The Woman Who Disputes". Similarly, changed "Al-Mumtahinah" to "Al-Mumtahanah" to match "The Woman to be examined".

Also, empty entries in the pdf are kept as before.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -322,8 +322,7 @@
   <string name="get_translations">Get Translations</string>
 
   <!-- help -->
-  <string name="email_us">If your question is not answered above, you can email quranandroid@gmail.com for support.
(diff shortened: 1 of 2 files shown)


ahmedre (approved):
جزاكم الله خيراً

PR #1852, opened 2022-01-22
Fix selection across multiple pages

ahmedre:
When selecting ayahs over multiple pages, the menu did not move to the
new page. This patch fixes it so now ayat can be selected on multiple
pages as was the case before. Note that there is a limit of 2 pages
worth of ayat (well, 4 if in tablet mode or side by side view, which
probably should be reduced in the future).


nacer80:
Still, this issue is not fixed when we choose large screens pages(madinah 1439).

PR #1859, opened 2022-01-26
Replace math function from Java's to Kotlin's

wiryadev:
Use Kotlin's math for `abs` and `min` function instead of Java's one


ahmedre (approved):

ahmedre:
jazakumAllah khairan

PR #1865, opened 2022-01-27
fix typo in variable name

wiryadev:
Change `tafaseer` to `tafseer` in one variable. I'm sorry if that's intentional and not a typo.


ahmedre (approved):
it was intentional (tafaseer is the plural of tafseer in Arabic), but will accept your change to be consistent. jazakumAllah khairan.

PR #1872, opened 2022-01-30
rewrite `QuranListAdapter` in Kotlin

wiryadev:
with minor change in `JuzListFragment` to filter null


ahmedre (approved):
jazakAllah khair

PR #1875, opened 2022-01-30
migrate `QuranHeaderPreference` to Kotlin

wiryadev:


app/src/main/java/com/quran/labs/androidquran/ui/preference/QuranHeaderPreference.kt
```diff
@@ -0,0 +1,45 @@
...
+    context: Context, attrs: AttributeSet, defStyleAttr: Int
+  ) : super(context, attrs, defStyleAttr) {
+    init()
+  }
+
+  constructor(
+    context: Context, attrs: AttributeSet, defStyleAttr: Int, defStyleRes: Int
+  ) : super(context, attrs, defStyleAttr, defStyleRes) {
+    init()
+  }
```
ahmedre, line 31: how about something like this instead:

```kotlin
class QuranHeaderPreference @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) : Preference(context, attires, defStyleAttr) {

   init {
      init()
   }
}
```

wiryadev: what about the constructor with 4 params? It isn't used, should we remove it?

ahmedre: true, please feel free to add defStyleRes: Int = 0 in the above and pass it to the `Preference` constructor to add it.

wiryadev: Change applied, i also moved code inside init function to the init block

ahmedre (approved):

ahmedre:
jazakumAllah khairan!

PR #1876, opened 2022-01-30
migrate `SeekBarPreference` to Kotlin

wiryadev:
with some adjustment to its subclasses


app/src/main/java/com/quran/labs/androidquran/ui/preference/SeekBarTextBrightnessPreference.kt
```diff
@@ -6,13 +6,15 @@ import android.util.AttributeSet
...
+  context: Context, attrs: AttributeSet
+) : SeekBarPreference(context, attrs) {
 
   override fun getPreviewVisibility(): Int = View.VISIBLE
 
-  override fun onProgressChanged(seek: SeekBar, value: Int, fromTouch: Boolean) {
-    super.onProgressChanged(seek, value, fromTouch)
-    val lineColor = Color.argb(value, 255, 255, 255)
-    mPreviewText.setTextColor(lineColor)
+  override fun onProgressChanged(seekBar: SeekBar?, progress: Int, fromUser: Boolean) {
```
ahmedre, line 15: is there any way we can make this seeker non-null here and in the other class?

wiryadev: Yes we can, pushed it just now

ahmedre (approved):
jazakAllah khairan!

PR #1888, opened 2022-02-12
Kotlinify AyahHighlight

aabdurrahman:


app/src/main/java/com/quran/labs/androidquran/ui/helpers/AyahHighlight.kt
```diff
@@ -0,0 +1,23 @@
+package com.quran.labs.androidquran.ui.helpers
+
+sealed class AyahHighlight(val key: String, val isTransition: Boolean) {
```
ahmedre, line 3: why not make this a `data` class, then you get `equals`, `hashCode`, and `toString` for free

PR #1908, opened 2022-03-04
Fix some warnings in BookmarksFragment

ahmedre:


app/src/main/java/com/quran/labs/androidquran/ui/fragment/BookmarksFragment.java
```diff
@@ -114,32 +116,27 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
-        item.setChecked(bookmarkPresenter.isShowingRecents());
-        return true;
-      }
-      case R.id.show_date: {
-        bookmarkPresenter.toogleShowDate();
-        bookmarksAdapter.setShowDate(bookmarkPresenter.isDateShowing());
-        item.setChecked(bookmarkPresenter.isDateShowing());
-        return true;
-      }
+    if (itemId == R.id.sort_date) {
```
ahmedre, line 119: will be broken on AGP 8 because R will become non-final

PR #1909, opened 2022-03-04
Add recitation feature

aabdurrahman:


app/src/main/java/com/quran/labs/androidquran/database/TranslationsDaoImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+class TranslationsDaoImpl @Inject constructor(
+  private val appContext: Context,
+  private val quranFileUtils: QuranFileUtils,
+) : TranslationsDao {
+
+  override suspend fun allAyahs(): List<QuranText> {
+    return withContext(Dispatchers.IO) {
+      val databaseHandler = DatabaseHandler.getDatabaseHandler(
+        appContext, QuranDataProvider.QURAN_ARABIC_DATABASE, quranFileUtils)
+      databaseHandler.getVerses(VerseRange(1, 1, 114, 6, -1), TRANSLATION)
```
ahmedre, line 25: this is going to be very memory heavy - can we just load what we need?

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
```diff
@@ -526,6 +529,27 @@ public void onPageSelected(int position) {
...
+    // Setup recitation (if enabled)
+    pagerActivityRecitationPresenter.bind(this, new PagerActivityRecitationPresenter.Bridge(
+        () -> numberOfPages,
+        () -> numberOfPagesDual,
+        () -> isDualPageVisible(),
+        () -> viewPager,
+        () -> audioStatusBar,
+        () -> ayahToolBar,
+        () -> slidingPagerAdapter,
+        () -> getSelectionStart(),
```
ahmedre, line 542: why do we expose all of these? can we not just share the existing models and avoid directly sharing ui items?

app/src/main/java/com/quran/labs/androidquran/view/AudioStatusBar.java
```diff
@@ -409,6 +453,83 @@ private void addDownloadProgress(@StringRes int text) {
     addView(ll, lp);
   }
 
+  @SuppressLint("NewApi")
```
ahmedre, line 456: is this for the image tint stuff? can we remove this suppress lint and make sure that either the model has the correct minSdk version or do an sdk check instead? otherwise can we use something from appcompat?

app/src/main/java/com/quran/labs/androidquran/view/AudioStatusBar.java
```diff
@@ -409,6 +453,83 @@ private void addDownloadProgress(@StringRes int text) {
...
+      addSeparator();
+      addSpacer();
+      addSeparator();
+      addButton(R.drawable.ic_transcript, false);
+      addSeparator();
+      addButton(recitationButton, R.drawable.ic_mic, false);
+    }
+  }
+
+  private void showRecitationStoppedMode() {
```
ahmedre, line 483: what do you think about swapping this out for a compose audio bar and toggle buttons based on state instead (or xml with show/hide calls? this was already a bit unwieldy and becomes troublesome if we keep trying to expand it like this

ahmedre (commented):
جزاكم الله خيراً

the pr is a bit long so difficult to properly reviewed but added some comments. will make another pass in sha' Allah after you update.

app/src/main/java/com/quran/labs/androidquran/view/AudioStatusBar.java
aabdurrahman: This `SupressLint` is not required. Removed.
(image tint is API 21+)

app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
aabdurrahman: Refactored it to reduce the exposed ui items. Removed `numberOfPages`, `numberOfPagesDual`, `viewPager` and `slidingPagerAdapter` Reduced to:
```kotlin
  class Bridge(
    val isDualPageVisible: () -> Boolean,
    val currentPage: () -> Int,
    val audioStatusBar: () -> AudioStatusBar?,
    val ayahToolBar: () -> AyahToolBar?,
    val ensurePage: (ayah: SuraAyah) -> Unit,
    val showSlider: (sliderPage: Int) -> Unit,
  )
```
* AudioStatusBar is needed to set audiobar.listener, recitation enabled mode, and switch audiobar mode
* AyahToolBar is needed to set recitation enabled mode

app/src/main/java/com/quran/labs/androidquran/database/TranslationsDaoImpl.kt
aabdurrahman: The entire range is needed, but the loaded list is only used in transient (not held), it's reduced to a single string that is a concatenation of the ayah texts.

app/src/main/java/com/quran/labs/androidquran/view/AudioStatusBar.java
aabdurrahman: Yes, that would be better. This is a much bigger change however, so may be better as a separate PR

ahmedre (commented):
جزاكم الله خيراً

ahmedre (approved):
جزاكم الله خيراً

PR #1917, opened 2022-03-12
show bookmark item as toggled when page is first entered

tahmid-23:
before when a page was bookmarked and you clicked on the bookmark, in the top bar it would show the bookmark as deactivated, this loads the bookmark cache state when the menu is opened

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
+++ b/app/src/main/java/com/quran/labs/androidquran/ui/PagerActivity.java
@@ -393,7 +393,7 @@ public class PagerActivity extends AppCompatActivity implements
     viewPager.setAdapter(pagerAdapter);
 
     ayahToolBar.setOnItemSelectedListener(new AyahMenuItemSelectionHandler());
-    viewPager.addOnPageChangeListener(new OnPageChangeListener() {
+    OnPageChangeListener onPageChangeListener = new OnPageChangeListener() {
 
       @Override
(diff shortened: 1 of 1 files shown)


ahmedre:
salam - jazakumAllah khairan -
i am not able to reproduce this issue - how are you seeing it? I bookmarked something and then went to the bookmarks tab and clicked it, and the page properly shows as bookmarked.

also, one question - why not also check split like the existing code?
jazakumAllah khairan.

tahmid-23:
it looks like this only happens on the 0th page because android won't call onPageSelected for page 0. I thought this happened on all pages. this will just call the handler manually

also isDualPagesVisible will not only check that it's not split screen, but also that we're not showing translations in split screen. previous code was prone to an error there too.

ahmedre:
onPageSelected should also be called for page 0, though not on the very first load - but I can't repro the bug you're trying to fix. the two things I've tried:

- bookmark an ayah in sura Fatiha. hit back. click the ayah from the bookmarks list. long press ayah and it correctly shows bookmarked state.
- bookmark page 1 and hit back. click page 1 from bookmarks. correctly shows bookmarked state.

tahmid-23:
1st page is surah al-ikhlas in the app code

ahmedre:
jazakAllah khairan - I can repro and I am seeing what you're seeing where page 0 isn't set if it's the page that the person directly jumped to

ahmedre (approved):
nice catch, jazakumAllah khairan

Issue #1177, opened 2019-05-06
Script to convert exported bookmarks to csv


ahmedre:
This was requested via email - but if we can provide a script (or a small website?) so that people can easily upload Quran for Android json files and have them output as csv instead would be useful.

murtraja:
@ahmedre , We could spin up a server that does this conversion, but I wanted to know, shouldn't this be done by the App itself? So, if I wanted to get CSV, I have to first export the bookmarks, then upload them and download the CSV. Instead, wouldn't it be more convenient if the App could export it in CSV itself in addition to JSON format?

ahmedre:
that's true - the only reason i didn't think to put it in the app directly is because i've only seen one person request this so far (and the default format is json which is easy to parse and i think should be shared with the iOS app (though not sure if it is or not)).

murtraja:
In that case, I can run a server on Heroku which does this. Do you have a language preference for the same?

I am asking this because in the future we might end up merging it with Quran.com backend.

Also, can you share the CSV format in which the output is expected (or should I plainly spit out CSV from JSON)?

ahmedre:
i don't think it has to be a service - i just want to point people to "use this to convert from json to csv format" if someone asks - so a simple script in any language would do the trick

murtraja:
@ahmedre, I checked on my iPhone5, I couldn't find any export bookmarks option. However the bookmarks tab correctly shows three kinds of bookmark - Page bookmarks, Ayah bookmarks and Recent pages. (there is no concept of tag)

ahmedre:
oh i didn't realize there aren't tags on the iOS app yet. i guess the hope is that the files would be shareable between both apps once the iOS app supports this.

MahmoudMabrok:
> i don't think it has to be a service - i just want to point people to "use this to convert from json to csv format" if someone asks - so a simple script in any language would do the trick

is it still open?

weekmo:
Is it still open?

ozbek:
Salam.

Yes, please. Issues with `PRs Welcome` tag are open to anyone wishing to contribute.

jibraniqbal666:
Salam @ahmedre bhai, I worked with you in Careem and i would love to contribute in this.
What if we add directly export CSV in the app?

ahmedre:
wa3laikum alsalam @jibraniqbal666 bhai,
nice to hear from you - sure, as long as the PR isn't in Flutter 😂 
[...]

PR #1946, opened 2022-04-16
#1177 Added new feature of export backup as CSV file

jibraniqbal666:
- Added CSV by just making everything by breaking line.
- Did some method extraction to avoid duplicate code.
- Schema will looks as below.

image

Schema:

Tags
--> List of Tags
Bookmarks
--> List of Bookmarks
RecentPages
--> List of Recent Pages.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/data/Constants.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/data/Constants.kt
@@ -84,6 +84,7 @@ object Constants {
   const val PREF_PAGE_TYPE = "pageTypeKey"
   const val PREF_IMPORT = "importKey"
   const val PREF_EXPORT = "exportKey"
+  const val PREF_EXPORT_CSV = "exportKeyCSV"
   const val PREF_LOGS = "sendLogsKey"
   const val PREF_DID_PRESENT_PERMISSIONS_DIALOG = "didPresentStoragePermissionDialog"
   const val PREF_WAS_SHOWING_TRANSLATION = "wasShowingTranslation"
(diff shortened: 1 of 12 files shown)


ahmedre:
jazakumAllah khairan Jibran - looks good!

2 questions -
1. if I try to import this as is into Google Drive, it might not work - what if we inline tags and add a type field so we can also inline recents? this way we just have a single table that's importable without needing to look up keys, etc? what do you think? something like:

```
type,sura,ayah,page,timestamp,tags
bookmark,1,,1,1650183444000,tag1|tag2|tag3
bookmark,2,1,2,1650190000,tag2
recent,15,,1650000000000,
```

in the above, I made a few tweaks from what you have -
I intentionally dropped ayahText - not needed, but have an ayah id that may or may not be there depending on whether it's an ayah bookmark or a sura bookmark. I changed ; to |, but not necessary, ; is fine. I also dropped the ids because they're irrelevant to someone exporting the data.

2. why the extra space before every new line?

jibraniqbal666:
> 1. if I try to import this as is into Google Drive, it might not work - what if we inline tags and add a type field so we can also inline recents? this way we just have a single table that's importable without needing to look up keys, etc? what do you think? something like:

1. I agree, we don't need tags as list as well, we can just bookmark and recent, i made this schema, so by just seeing the data class, right we should normalize it as recent and bookmark, i will make the changes.
2. just for the readability as i believe it will trimmed down if you import it in proper spreadsheet viewer, i can do trim before writing. :)

jibraniqbal666:
New schema:
image

ahmedre (approved):

ahmedre:
jazakAllah khair!

ibardawil:
Assalamu alaikum,
import csv file doesn't work

ahmedre:
this feature is just for export - there's an import/export for json format that you should use if you want to import.

PR #1951, opened 2022-04-20
hide margins if no overlay text

benomaire:


app/src/main/java/com/quran/labs/androidquran/view/HighlightingImageView.java
```diff
@@ -525,6 +522,15 @@ protected void onDraw(@NonNull Canvas canvas) {
...
     }
 
+    if(!didDraw){
+      setPadding(horizontalSafeOffset, topSafeOffset, horizontalSafeOffset, bottomSafeOffset);
+    } else {
+      setPadding(horizontalSafeOffset,
+          topBottom + topSafeOffset,
+          horizontalSafeOffset,
+          topBottom + bottomSafeOffset);
+    }
```
ahmedre, line 532: jazakumAllah khairan - may we move this back into `setIsScrollable` with a check? this `onDraw` method gets called every frame, so we want to keep it as light as possible - whereas `setIsScrollable` just happens once

benomaire: I am not sure, but I don't see an easy way to move this block back to `setIsScrollable`, because `setIsScrollable` is not aware of what happens in `onDraw`

Some observations:

- show pages 1 and 2 (landscape, dual page) for the first time, then `setIsScrollable` is executed 8 times and `onDraw` is executed 6 times.
- swipe to pages 3 and 4, then `setIsScrollable` is executed 2 times and `onDraw` is executed 4 times (reduces to 2 times when this `if` block is removed).
- swipe back to pages 1 and 2, then only `setIsScrollable` is executed twice.

Issue #1963, opened 2022-05-02
app: copied translation does not have the surah:ayath reference


nas5k:
Ealier, the copied text when pasted looked like this:
===
(Dr. Mustafa Khattab)
Indeed, there was a group of My servants who used to pray, ‘Our Lord! We have believed, so forgive us and have mercy on us, for You are the best of those who show mercy,’

-Surah Al-Mu'minun, Ayah 109
===

Now, the surah/ayah reference is missing, and it looks like this:
===
Dr. Mustafa Khattab
Indeed, there was a group of My servants who used to pray, ‘Our Lord! We have believed, so forgive us and have mercy on us, for You are the best of those who show mercy,’
===

PR #2006, opened 2022-07-02
Show Ayah Reference when Arabic is not Shown

DoozyDoz:
In response to issue #1963

before changes
Screenshot_20220702_103151
after changes
Screenshot_20220702_103123

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/util/ShareUtil.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/util/ShareUtil.kt
@@ -14,6 +14,7 @@ import com.quran.labs.androidquran.common.QuranAyahInfo
 import com.quran.labs.androidquran.data.QuranDisplayData
 import com.quran.labs.androidquran.model.translation.ArabicDatabaseUtils
 import com.quran.labs.androidquran.ui.util.ToastCompat
+import com.quran.labs.androidquran.ui.util.TypefaceManager
 import dagger.Reusable
 import java.text.NumberFormat
 import java.util.Locale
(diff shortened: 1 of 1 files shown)


ahmedre:
wa3laikum alsalam,
jazakumAllah khairan - however, the sura string is already there, just under the arabic - by adding it here we'd have it multiple times which would be too much. we moved it there to support cases where multiple translations are shown, and thus showing this multiple times wouldn't be ideal.

DoozyDoz:
> wa3laikum alsalam, jazakumAllah khairan - however, the sura string is already there, just under the arabic - by adding it here we'd have it multiple times which would be too much. we moved it there to support cases where multiple translations are shown, and thus showing this multiple times wouldn't be ideal.

Salam alaikum brother
I noticed too, its only abscent when the option to not show ayah is unchecked in setting( ie. If only translation is shown)
So I guess I should check the option before appending the surah reference

app/src/main/java/com/quran/labs/androidquran/util/ShareUtil.kt
```diff
@@ -77,6 +78,12 @@ class ShareUtil @Inject internal constructor(private val quranDisplayData: Quran
           append(text)
         }
       }
+      val arabicIsShownBeforeTranslation = QuranSettings.getInstance(context).wantArabicInTranslationView()
+      if (!arabicIsShownBeforeTranslation){
```
ahmedre, line 82:
```suggestion
      if (ayahInfo.arabicText == null){
```

ahmedre:
jazakumAllah khairan - will merge this shortly in sha' Allah.

ahmedre:
jazakumAllah khairan

PR #2008, opened 2022-07-03
Minor cleanup of AudioUtils

ahmedre:


app/src/main/java/com/quran/labs/androidquran/presenter/audio/AudioPresenter.kt
```diff
@@ -100,7 +100,7 @@ constructor(private val quranDisplayData: QuranDisplayData,
           context.getString(R.string.highlighting_database))
     } else if (gaplessDb != null && !File(gaplessDb).exists()) {
       getDownloadIntent(context,
-          audioUtil.getGaplessDatabaseUrl(qari)!!,
+          getGaplessDatabaseUrl(qari)!!,
```
M-KH, line 103: The app will crash if getGaplessDatabaseUrl() returns null. Why don't you make it optional?

PR #2011, opened 2022-07-05
[feat] share ayah as audio 

DoozyDoz:
This is an attempt to address issue #1731
### Screenshot
Screenshot_20220705_183433_com quran labs androidquran debug

### Addressed
- Added option to share audio ayah after its been highlighted.
- Share ayat from different surah as single audio file.
- Changed the "share ayah text" icon.
### Not Addressed
- Asking users to download audio files in case they are missing.
- There is no functionality to add isthaatha before non-beginner ayat
- When sharing audio to WhatsApp it sometimes fails.
- Code is poorly written and I welcome any advice

PR #2072, opened 2022-09-11
feat share audio

ahmedre:
- Feat share ayah as audio (#1)
- manifest cleanup
- cleanup
- Icons changed
- proper "sharable audio" name
- share intent title
- [bugfix] - app was crushing when full surah is selected
- [bugfix] - app was crushing when full surah is selected 2
- request user to download files if not present
- clean up
- transffered to audio feature
- removed greek prefix
- check cache directory existance at sharing
- Move audio sharing to a separate feature module
- Initial clean up to PagerActivity for audio share
- Move some methods into audio file utilities
- Use normal application cache directory


ahmedre:
This is @DoozyDoz's pull request from #2011, but continued here so I can have write access. I've done a set of fixes and moved some things around, but it's not ready yet.

ahmedre:
Things that are left before merging this PR:
- [ ] better handling for cases when the audio isn't downloaded (today, it requests download, but doesn't try to share afterwards - we should either have proper messaging or fix this).
- [ ] sharing is only for gapless audio today
- [ ] the majority of the code added to `PagerActivity` should move out
- [ ] cleanup of `AudioShareUtils`

ahmedre:
A bigger question though is whether or not we should continue with this PR and merge it in the first place - while the above issues can be worked through and resolved, there remains one major issue that's difficult to work around - and that is the audio timings are inherently inaccurate. This can easily be tested by sharing a single ayah - it feels unnatural unless the timings are precise.

My worry is that introducing this feature will greatly increase support requests due to inaccurate timings causing shared audio to be cut.

curious to what you think @nacer80 / @benomaire

ahmedre:
One more question - today, this code supports sharing across suras (i.e. share a file between sura 113 ayah 1 through the end of sura 114) - if we are to continue with this, I am thinking we should restrict this to sharing within one sura only. In addition to making the code simpler, guessing in most cases, it makes sense to share an ayah or set of ayat from one sura rather than a range across suras.

nacer80:
We will faced tow issues some timing:

1- if the recitation of Ayah shared doesn't stop in the end of Ayah (Sheikh  recited tow Ayahs, so the there is letter between the end of Ayah and the beginning of the second)
2- if timing doesn't accurate.

DoozyDoz:
> One more question - today, this code supports sharing across suras (i.e. share a file between sura 113 ayah 1 through the end of sura 114) - if we are to continue with this, I am thinking we should restrict this to sharing within one sura only. In addition to making the code simpler, guessing in most cases, it makes sense to share an ayah or set of ayat from one sura rather than a range across suras.

Same thoughts but I think it's okay since it includes basmala/isthaadha where necessary. The resultant audio file is simply like 2 surahs being played one after the other which is normal, right?

DoozyDoz:
> A bigger question though is whether or not we should continue with this PR and merge it in the first place - while the above issues can be worked through and resolved, there remains one major issue that's difficult to work around - and that is the audio timings are inherently inaccurate. This can easily be tested by sharing a single ayah - it feels unnatural unless the timings are precise.
> 
> My worry is that introducing this feature will greatly increase support requests due to inaccurate timings causing shared audio to be cut.
> 
> curious to what you think @nacer80 / @benomaire

Is there any way I can be of help in this regard? how can one update the timing data if they wish to? what tools were used to time the initial set of gapless timings?

ahmedre:
we can at least enable it for one qari - so your sharing is always shared with one qari whose timings are pretty accurate - one of the slower qaris probably. will try to resume this at some point in the future in sha' Allah.

PR #2202, opened 2023-02-22
Remove remaining gapless strings in Arabic

ahmedre:


ahmedre:
thanks to @nacer80 for the report, جزاكم الله خيراً

PR #2207, opened 2023-02-25
Show footnotes inline

ahmedre:
Previously, footnotes were being removed. This patch updates it so that
footnotes are added inline while reading instead.


ahmedre:
light
dark

ahmedre:
this is a similar approach to what @mohamede1945 did in quran/quran-ios. the nice thing about it is it's simpler. I spent a lot of time thinking about this but unsure about the ux for other approaches (ex having a superscript and clicking shows the footnote - in this case, where do we show it? bottom sheet? dialog? what if the footnote is super short? super long? etc).

Issue #2286, opened 2023-05-23
App trackers found on the Quran app.


Taratect:
Are the developers aware that the app contains google trackers?
https://reports.exodus-privacy.eu.org/en/reports/com.quran.labs.androidquran/latest/

I'm terribly put off by knowing the fact that Google has ability to know when I read the words of my Lord which is is supposed to be extremely personal is being used to make money off selling to data brokers at the expense of my privacy.

I really liked the app and Frankly I'm disappointed.
So, I'd really appreciate it if you guys would remove such intrusive codes from your app. These type of trackers has no place in apps used for prayers and duas

ahmedre:
Salam 3alaikum,

as the link you sent mentions, we use two Google services:
- Crashlytics
- Firebase Analytics

We use Crashlytics to report crashes - anytime the application crashes, it lets us know that a crash happened, along with where in the code it happened. None of the developers have any way to tie this crash to a particular person. It literally is just the part of our code (line number and class name) where the crash occurred, plus any generic metadata we add to it.

We use Firebase Analytics to gather anonymous analytics on things like "how many people enable this setting" or "how many people use translation view in line versus the full screen translation view" in order to make decisions (if I remove this feature, will people be really upset? or I expect more people would be using this feature, why aren't people using it? is there a problem or does it need to be more obvious?) Again, even in this case, we have no way of tying this back to you. I get aggregate numbers.

Both Crashlytics and Firebase Analytics give developers an option to "tie data to a userid" for example - so applications that require a log in typically use this and can link this information back to you personally (i.e. "I know person with email x@y.com clicked this link and used the app for z minutes," etc). We don't do that.

We've thought about replacing Crashlytics with Bugsnag at some point, but even if we did so, it'd still show up as a tracker, even though again we're just using it for crash reporting.

In summary, we're building this app for the sake of Allah - not to profit from you directly or indirectly (i.e. there's no "this app is free but we make money from ads so the more ads you click the more money we make), and we don't gather user data so we don't have anything to sell. Everything we collect is anonymous for the purposes of making the app better and I can't tie it back to you.

If this bothers you, please consider building from source, we've provided a flag to disable Firebase for those cases, or use the version from F-Droid, since that one has those disabled.

jazakumAllah khairan.

Taratect:
[...]

PR #2300, opened 2023-06-13
IzzyOnDroid in README.md

KingLucius:
- Adding IzzyOnDroid Badge.
- Add available version on IzzyOnDroid shield.

[Edit]
It's usefull for devices without play store and depend on external stores like f-droid

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,8 @@
 [![Build Status](https://github.com/quran/quran_android/actions/workflows/build.yml/badge.svg)](https://github.com/quran/quran_android/actions/workflows/build.yml)
 [![Version](https://img.shields.io/github/v/release/quran/quran_android?include_prereleases&sort=semver)](https://github.com/quran/quran_android/releases/latest)
 [![Github Downloads](https://img.shields.io/github/downloads/quran/quran_android/total?logo=Github)](https://github.com/quran/quran_android/releases)
-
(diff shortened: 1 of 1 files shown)


benomaire:
If IzzyOnDroid is just copying the apk from github, then the inclusion of this badge is a choice to be made with no preference on my side.
On one hand, IzzyOnDroid is an alternative to f-droid for many apps not on the main f-droid repo, 
On the other hand, I think most users who know IzzyOnDroid can find the app there without this badge.

benomaire:
BTW, might be useful at this stage to build another apk without the "Google trackers" and list it in the releases, so we have 2 apks like: 

- quran-3.3.3.apk
- quran-no-analytics-3.3.3.apk

then if the "no-analytics" version is hosted on IzzyOnDroid, the privacy flags will be off for users who get the apk from there. 
Again this is just a convenience for users of IzzyOnDroid who will get automatic updates and an apk with no "trackers" without manually checking github releases.

I guess that will also resolve #2286

IzzySoft:
* If you want me to check that no-analytics APK beforehand, just let me know where to find it
* as for the badge: it would help those who do not (yet) know. Otherwise you could argue the same about the PlayStore and Github badges :zany_face:

MahmoudMabrok:
@KingLucius  I think badge is already added, if yes, could we close current PR ?

KingLucius:
> @KingLucius  I think badge is already added, if yes, could we close current PR ? 

Ok

PR #2305, opened 2023-06-17
Make footnotes collapsed by default

ahmedre:
Add logic to collapse all footnotes by default, and allow clicking a
footnote to expand it.


ahmedre:
collapsed:
collapsed

one footnote expanded:
expanded

before this change, all footnotes were always expanded.

PR #2320, opened 2023-07-15
Update Kotlin to 1.9.0

ahmedre:
Combine lint options for all Android projects to share Lint exclusions.


common/ui/core/build.gradle
```diff
@@ -1,5 +1,4 @@
 plugins {
-  id 'quran.android.library.android'
   id 'quran.android.library.compose'
```
ahmedre, line 2: only need one of these since they overlap in their functionality

PR #2351, opened 2023-08-20
Target sdk 34

ahmedre:


ahmedre:
depends on robolectric/robolectric#8404

PR #2391, opened 2023-09-24
Add some more qaris

ahmedre:


ahmedre:
adds:
- Hady Toure
- Nabil Ar-Rifai
- Noreen Siddiq
- Tawfeeq as-Sawaigh

makes:
- Dr. Ayman Suwaid as Gapless

prepares sheikh Khalid Jalil but doesn't add him yet until we work out some issues with the recitation.

PR #2438, opened 2023-10-23
Generalize Portuguese strings and fix Malaysian

ahmedre:
Some Malaysian strings were still set as Burmese from when we were using
values-my instead of values-ms. This renames them. Also, keep the
Portuguese folder for all Portuguese languages instead of just for
Brazil.


ahmedre:
jazakumAllah khairan @mohamede1945 for the heads up on this one.

PR #2515, opened 2023-12-24
Improve translations view

ahmedre:
Fix the case where there are no translations, and apply minor fixes.


ahmedre:
before:
before

after:
after

PR #2539, opened 2024-01-12
Fix and catch some crashes

ahmedre:


app/src/main/java/com/quran/labs/androidquran/service/QuranDownloadService.java
```diff
@@ -239,13 +239,15 @@ private void sendNoOpMessage(int id) {
 
   @Override
   public int onStartCommand(Intent intent, int flags, int startId) {
-    // if it's a download, it wants to be a foreground service.
-    // quickly start as foreground before actually enqueueing the request.
-    if (ACTION_DOWNLOAD_URL.equals(intent.getAction())) {
-      notifier.notifyDownloadStarting();
-    }
+    if (intent != null) {
```
ahmedre, line 242: intent can be null if the service restarts, since we return `START_NON_STICKY`

PR #2571, opened 2024-01-31
Integrate audio bar

ahmedre:


ahmedre:
delta in size is due to material-icons-extended, which should mostly clean up when we build a release build with r8.

PR #2628, opened 2024-03-17
Fix required downloads for audio

ahmedre:
After some of the required downloads complete, the audio was not
starting, due to needing more required downloads. This is because the
download bridge missed the FileDownloaded event and only was handling
the batch download success event.


ahmedre:
closing, need to rethink this.

PR #2643, opened 2024-03-25
Fix: Scroll state in retained in the partial translation mode

### LINKED ISSUE #2598: Retain scroll state in partial translation mode
I can repro this on my phone if I completely go back to home and come back into the app.
https://github.com/quran/quran_android/assets/47771/a3225c30-39fc-487d-828d-13eec5c56bb9

ikerfah:
Fixes #2598 [Retain scroll state in partial translation mode](https://github.com/quran/quran_android/issues/2598)

The issue was with the call to `scrollTo(0, 0)` within the `setAyahs()`, so basically when the Ayahs are set the scroll is rested to 0,0.

Removing `scrollTo(0, 0)` persists the scroll position since `InlineTranslationView` extends `ScrollView` and it saves the scroll state out of the box. but this introduce a new **problem** which is when you select a different ayah then the scroll will not be rested. to fix this I added a flag called `ayahHasBeenChanged` to reset the scroll to 0,0 when ayah is changed

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.kt
@@ -28,6 +28,7 @@ class InlineTranslationPresenter @Inject constructor(
 ) {
   private val scope = MainScope()
   private var cachedTranslations = emptyList<LocalTranslation>()
+  private var lastVerseRange: VerseRange? = null
 
   init {
     translationListPresenter.translations()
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.kt
```diff
@@ -38,11 +38,11 @@ class InlineTranslationPresenter @Inject constructor(
       .launchIn(scope)
   }
 
-  suspend fun refresh(verseRange: VerseRange) {
+  suspend fun refresh(verseRange: VerseRange, ayahHasBeenChanged: Boolean) {
```
ahmedre, line 41: if we keep track of the last verse range within `InlineTranslationPresenter`, we can avoid having to update all the callers to pass in whether or not the ayah has changed - can we do that instead? i.e.:

```kotlin
private var lastVerseRange: VerseRange? = null

suspend fun refresh(verseRange: verseRange) {
  val ayahHasBeenChanged = verseRange == lastVerseRange
  // existing logic here
  lastVerseRange = verseRange
}
```

ahmedre (commented):
jazakumAllah khairan, really appreciate the great work on this and looking forward to merging it! just one small comment in the PR that I think will simplify the PR a bit.

app/src/main/java/com/quran/labs/androidquran/presenter/translation/InlineTranslationPresenter.kt
ikerfah: Looks good! I changed the PR (forced pushed to emit the old commit)

ahmedre (approved):
jazakumAllah khairan!

PR #2648, opened 2024-03-29
Android Auto Support  🚗 

jibraniqbal666:
Salam Y'all

Alhamdullilah, I was finally able to add the android auto support but with limitations, and the code is pretty basic for now.

Video (With sound):

https://github.com/quran/quran_android/assets/30576574/b4790b57-ebba-459d-880e-ada76fe25129

Feature:
- Surah List
- Media Play

Limitation:
- Only one Sheikh.
- No Pause or Seek Support.

Most of the code is in separate module and I am planning to add more features later, inshaAllah

Run with `autoquran` configuration to try it with Automotive emulator.


ahmedre:
wa3laikum alsalam,
masha'Allah Jibran this looks really promising! Looking forward to following your progress here. I am guessing at some point we should probably make our existing `AudioService` conform to some of these interfaces (`MediaBrowserServiceCompat`) so they work with Android Auto out of the box.

jibraniqbal666:
@ahmedre, should we merge this or i should continue working on this branch?

ahmedre:
let's keep working on it until it's in a more ready state?

jibraniqbal666:
Added Next Previous & Seek

jibraniqbal666:
[Update]

https://github.com/user-attachments/assets/041beae3-8b0e-4975-92dc-ab2dbb2e4fb5


Salam, @ahmedre, I have added basic media player functionality, Just required clean up.
Do you think we should merge it after that.

Here is the state below.
Feature:
Surah List ✅ 
Media Play ✅ 
Pause ✅ 
Next ✅ 
Previous ✅ 
Seek ✅ 

Limitation:
Only one Sheikh  ❌

ahmedre:
wa3laikum alsalam Jibran,
jazakumAllah khairan for the great work! agree with you - will aim to review this weekend in sha' Allah and then we can merge it.

I haven't checked the code yet, but for the sheikh, can we change it from within the app on the phone itself (i.e. the last used reciter?) I think if so, that'd save us from needing to do a proper sheikh selection feature for some time.

jazakumAllah khairan!
walsalam 3alaikum.

jibraniqbal666:
[Update]
It is now cleaned up.

Here is the state below.
Feature:
Surah List ✅
Media Play ✅
Pause ✅
Next ✅
Previous ✅
Seek ✅
2 Shaikh (Gapeless) ✅  

Limitation:
Non-Gapeless Shaikh ❌

jibraniqbal666:
> wa3laikum alsalam Jibran, jazakumAllah khairan for the great work! agree with you - will aim to review this weekend in sha' Allah and then we can merge it.
> 
> I haven't checked the code yet, but for the sheikh, can we change it from within the app on the phone itself (i.e. the last used reciter?) I think if so, that'd save us from needing to do a proper sheikh selection feature for some time.
> 
> jazakumAllah khairan! walsalam 3alaikum.

did not try this, I only have emulator to test, so i am only testing in automotive emulator and there is only one available on for arm64, and it believe it doesn't require phone.

autoquran/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
ahmedre, line 1: nit - remove this class

autoquran/src/androidTest/java/com/quran/labs/autoquran/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.Assert.*
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
```
ahmedre, line 17: nit, remove this class

autoquran/src/main/java/com/quran/labs/autoquran/QuranAudioService.kt
```diff
@@ -0,0 +1,99 @@
...
+import com.quran.labs.autoquran.common.MediaSessionCallback
+import com.quran.labs.autoquran.common.SurahBuilder
+import com.quran.labs.autoquran.common.filterWithQari
+import com.quran.labs.autoquran.di.DaggerServiceComponent
+import javax.inject.Inject
+
+class QuranAudioService : MediaBrowserServiceCompat() {
+  @Inject
+  lateinit var surahBuilder: SurahBuilder
+  private lateinit var mSession: MediaSessionCompat
```
ahmedre, line 19: nit - `session`?

autoquran/src/main/java/com/quran/labs/autoquran/common/SurahUtils.kt
```diff
@@ -0,0 +1,39 @@
+package com.quran.labs.autoquran.common
+
+import android.content.Context
+import android.text.TextUtils
+import com.quran.data.core.QuranConstants
+import com.quran.labs.autoquran.R
+
+fun getSuraName(
```
ahmedre, line 8: are we able to use the method in `QuranDisplayData`?

autoquran/src/main/java/com/quran/labs/autoquran/di/ServiceModule.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.quran.data.page.provider.QuranDataModule
+import com.quran.data.source.PageProvider
+import com.quran.data.source.QuranDataSource
+import com.quran.labs.autoquran.QuranAudioService
+import dagger.Component
+import dagger.Module
+import dagger.Provides
+
+@Module
+class ServiceModule {
```
ahmedre, line 12: can we not use the existing module for this

autoquran/src/main/res/values/sura_names.xml
```diff
@@ -0,0 +1,238 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ahmedre, line 1: can we not share the existing one? I understand because it's in `app` now, but we can move it (and the icons) to a `common` module and share it?

autoquran/src/test/java/com/quran/labs/autoquran/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
+package com.quran.labs.autoquran
```
ahmedre, line 1: remove class

ahmedre (commented):
code looks good, thanks - I tried this on an actual Android Auto device and didn't see the icon present in the list of apps. From my understanding, this is still shipped as a single apk, right? am I missing something / any ideas for why it isn't appearing 🤔

ahmedre:
> did not try this, I only have emulator to test, so i am only testing in automotive emulator and there is only one available on for arm64, and it believe it doesn't require phone.

- no need to handle non-gapless sheikh
- I guess the idea though is most people would have the app on the phone and then would want to see it on the car (as opposed to having a separate standalone app - but as mentioned in the other comment, I can't see this on an actual device

ahmedre:
let me give this a shot actually and see if I can do something based on your PR

jibraniqbal666:
> this

try to run on autoquran config not normal app

autoquran/src/main/java/com/quran/labs/autoquran/di/ServiceModule.kt
jibraniqbal666: No as i believe we run it as different application.

ahmedre:
yes - that's for the Android Auto OS standalone apk (which works). I wasn't able to make it show up on an actual device until I enabled developer settings within Android Auto. 

I made some changes, please see #2891.

jibraniqbal666:
> > did not try this, I only have emulator to test, so i am only testing in automotive emulator and there is only one available on for arm64, and it believe it doesn't require phone.
> 
> * no need to handle non-gapless sheikh
> * I guess the idea though is most people would have the app on the phone and then would want to see it on the car (as opposed to having a separate standalone app - but as mentioned in the other comment, I can't see this on an actual device

i don't have emulator carplay which run as extension of phone, maybe someone can try on actual car. (i don't have car either :))))

autoquran/src/main/java/com/quran/labs/autoquran/common/SurahUtils.kt
jibraniqbal666: It requires more work to use this and have to move all the surah_name and stuff. i believe this would be much better in different PR.

autoquran/src/main/java/com/quran/labs/autoquran/di/ServiceModule.kt
jibraniqbal666: I believe due to different application of autoquran it require new graph of app and i don't wanna bload by adding everything.

autoquran/src/main/res/values/sura_names.xml
ahmedre: I already did this in my branch.

autoquran/src/main/java/com/quran/labs/autoquran/di/ServiceModule.kt
ahmedre: I already did this in my branch.

ahmedre:
will go ahead and merge this, since it requires publishing a separate apk anyway to the Play Store to work, so we can merge it while we iron out the remaining issues in the other PR in sha' Allah.

jazakAllah khair for the awesome work on this!

Issue #1553, opened 2021-02-26
Foldables


ahmedre:
Try on a foldable emulator to see what low hanging fruit / wins there are - might just be a matter of allowing dual page in portrait mode on these devices.

amirian:
no plan to display two pages on foldable devices?

ahmedre:
Fixed by #2670 - جزاكم الله خيراً @kadirkid

PR #2670, opened 2024-04-21
Add support for dual pages on foldable devices

kadirkid:
### Goal
Support dual pages on foldable devices while the orientation is still portrait. This will only be supported on foldable devices that fold horizontally.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/di/module/activity/PagerActivityModule.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/di/module/activity/PagerActivityModule.kt
@@ -33,7 +33,7 @@ object PagerActivityModule {
   @Provides
   @ActivityScope
   fun provideImageWidth(@ActivityContext context: Context, screenInfo: QuranScreenInfo): String {
-    return if (QuranUtils.isDualPages(context, screenInfo)) {
+    return if (QuranUtils.isDualPages(context, screenInfo, false)) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/quran/labs/androidquran/di/module/activity/PagerActivityModule.kt
```diff
@@ -33,7 +33,7 @@ object PagerActivityModule {
   @Provides
   @ActivityScope
   fun provideImageWidth(@ActivityContext context: Context, screenInfo: QuranScreenInfo): String {
-    return if (QuranUtils.isDualPages(context, screenInfo)) {
+    return if (QuranUtils.isDualPages(context, screenInfo, false)) {
```
kadirkid, line 36: I have no way of injecting this value here so ignoring it for now. @ahmedre Could you elaborate a bit more on the use case here?

ahmedre:
JazakumAllah khairan @kadirkid  - this fixes #1553.

app/src/main/java/com/quran/labs/androidquran/di/module/activity/PagerActivityModule.kt
ahmedre: discussed offline, moving forward, will consider moving this check into `QuranScreenInfo` if possible.

ahmedre (approved):

PR #2891, opened 2024-10-06
Improve Android Auto Support

ahmedre:
This patch builds on #2648, and makes a few changes - namely:
- support for vanilla Android Auto (not just Android Auto OS)
- shared code between vanilla Android Auto and Android Auto OS

The "vanilla Android auto" ships with the APK and is available on devices that use Android auto on the phone. The "autoquran" one is based on the "autoquran" target (instead of app) and is a completely separate apk for Android Auto OS devices.


ahmedre:
cc @jibraniqbal666 please take a look.

some notes - on both of our branches, audio doesn't play back for me in the Android Auto OS version.
I also haven't successfully made it run on vanilla Android Auto yet (the icon shows up, but when I launch, nothing happens - need to check since it might be fixed with these recent changes)

ahmedre:
## Blockers
- [ ] fix audio playback

## Nice to Have
- [ ] support setting a Qari from settings (or use the default qari) instead of just the first 2 qaris
- [ ] play the file back from the local file system if it's downloaded instead of always getting it from the web
- [ ] make the ui nicer on Android Auto (perhaps copy what's in Android Auto OS of the auto-coloring of boxes)

jibraniqbal666:
> tested - notes:
> 
> * on Android Auto (not Android Auto OS), the boxes don't have a background color so they're kind of dull looking, unlike the nice screenshots from Android Auto OS
> * audio playback doesn't work yet (actual Android Auto, and emulator for Android Auto OS) - it seems like we need to implement audio attributes and/or handle ducking, etc, since, I noticed that if I try to do something (ex ok google), it'll actually play a tiny bit of the sura while it transitions it in and out.

How did you debug the Android Auto? Is there emulator? I would like to fix these issue :)))

jibraniqbal666:
> cc @jibraniqbal666 please take a look.
> 
> some notes - on both of our branches, audio doesn't play back for me in the Android Auto OS version. I also haven't successfully made it run on vanilla Android Auto yet (the icon shows up, but when I launch, nothing happens - need to check since it might be fixed with these recent changes)

Which emulator are you using for Android Auto OS, or is it real device?

ahmedre:
I actually have an Android Auto device so I was able to test on a real device - however, I was planning on testing the emulator also for vanilla Android auto - I found [this](https://stackoverflow.com/questions/76482834/can-we-test-android-auto-purely-in-emulators-2023) - tldr you launch the Android Auto OS emulator, and then you launch a vanilla Android emulator and install Android Auto on it and connect them.

ahmedre:
I just migrated the auto stuff to media3, which simplifies a lot of things - it does a lot of the ordinary things for us (handling playback, pause, updating the time, next, previous, ... - things you were doing manually before). This one I tested using the method from the Stackoverflow post linked above for testing and it works (whereas the version before that didn't work) - and this one also works on an actual Android Auto device.

It doesn't show the grid view anymore, but can take a look at that in the future in sha' Allah. Presumably, I think I can also make it support all the Qaris in an efficient way in sha' Allah, but need to give it a shot.

I also didn't test the standalone Android Automotive app yet, but not really too concerned about that one since I am more interested in Android Auto support (anyone with an Android phone + a device that supports Android Auto + the Android Auto app) rather than Android Automotive support (a completely separate app, needs an Android Automotive device, etc).

PR #2937, opened 2024-11-06
refactor: reorder step of build

MahmoudMabrok:
If PR has issues related to lint or test, so after fix we will run build step again with new push event, so I see we can postpone build step which take time and resources until we make sure PR is fine as code quality checks.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,9 +27,6 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
 
-      - name: Build debug APK
-        run: ./gradlew assembleMadaniDebug
-
       - name: Run lint
         run: ./gradlew lintMadaniDebug
 
@@ -52,6 +49,9 @@ jobs:
         with:
           name: unit_test_report
           path: app/build/reports/tests/testMadaniDebugUnitTest/
+
+      - name: Build debug APK
+        run: ./gradlew assembleMadaniDebug
 
       - name: Download Previous Debug APK
         uses: dawidd6/action-download-artifact@v6


ahmedre:
Interesting idea - what if, instead, we split lint out into a different machine, so that both builds run in parallel? This would potentially speed things up for successful builds, and would notify folks of lint issues quicker.

MahmoudMabrok:
for sure, we can do them in parallel (which also will add overhead of new checkout and install dependencies ) but we need to get failure quickly plus we do not need to waste time of build while lint is fail, so with new push that fix lint issues we will build again.

so we would make them sequential to save time of checkout and installing, plus with worst case of lint issue we do not waste time of runners.

ahmedre:
I think we should keep the order the way it is already, because the compiled code is needed for tests (and, if I am not mistaken, for some lint rules, even if the apk itself isn't required). It also makes more sense. To compare two PR runs around the exact time - one of this PR and one of another:

- [this PR](https://github.com/quran/quran_android/actions/runs/11712452534/job/32776990938?pr=2937)
* Lint: 3 minutes 55 seconds
* Tests: 38 seconds
* Build: 1 minute 49 seconds

- [today](https://github.com/quran/quran_android/actions/runs/11767824846/job/32776946524?pr=2939) - PR run around the same time as the above one.
* Build: 1 minute 51 seconds
* Lint: 1 minute 58 seconds
* Tests: 36 seconds

I am ok extracting lint into a parallel step (along with tests, perhaps), but I don't think it'll save us much in terms of time today, considering we do have to build.

MahmoudMabrok:
I got the point now, but what I was trying to save is time of cases that pr has issues with lint or tests, in these case we will save time for build the app, but seems with current case build step is making overall time is lesser.  

To be noted, might some caching make numbers not exact or ready to take as benchmark. 

at final, we can stick with current implementation (without PR).

ahmedre:
جزاكم الله خيراً

PR #2939, opened 2024-11-10
Improve Android Auto app

ahmedre:
- Implement search (qari only for now)
- Implement basic playlist support (selecting a sura puts the mushaf for
  that qari as a playlist positioned at the chosen sura)
- Make qari a top level category to allow for adding all qaris


ahmedre:
note - I tested removing the limitation of 2 qaris and it works fine now. I left the limit as it was so that we don't ship with all qaris until we implement recent playback support (since, otherwise, it'd be too much of a distraction for people driving a car). Also, grid support needs some sort of album art image per qari, since, right now, it looks very empty. In worst case, we can make everything use list support until we implement this, but ideally, we should do something here.

ahmedre:
cc @jibraniqbal666

Issue #1086, opened 2019-02-17
Android Auto support


ahmedre:
Let's consider adding support for Android Auto.

MonikaJethani:
@ahmedre Can I work on this?

ahmedre:
@MonikaJethani please feel free

benomaire:
@MonikaJethani Any updates?

farooqu:
I think it would be nice to consider the desired feature set for Android Auto since it obviously would not be the full app experience.

jibraniqbal666:
I wanna work on this next when i will have some time. @ahmedre :)
- I am thinking of maybe adding ayah by ayah screen, which is recited by desired reciter.
- wanna do it in compose :).

ahmedre:
sure - but before doing so, let's figure out what it will look like and what it will do just so we're aligned.
btw, we are using compose, it's just all hidden in the closed source portion of the repo (the new page rendering is all done in compose today).

jibraniqbal666:
yeah let's do that, how do we proceed for new feature, which needs design.
what do you mean by closed source portion?

ahmedre:
historically, people clone the repo, add some ads on top, and publish as is (keeping our servers, our email contact info, etc). as a result, we get a lot of complaints and so on. to make this a little more difficult, we've started keeping parts of the app in a closed source repo - the open source repo is still the base, but some code isn't there.

the intention longer term is to have a set of libraries that are open sourced that apps can build a Quran app on top of, instead of a full app like is the case today.

with respect to auto - I mean let's think about exactly what we want to do and support (feature wise even not code wise). I imagine something like this would be focused around the audio experience, but curious what you had in mind. I also would think there are probably higher priority issues than this if you're interested can point you to some of them.

jibraniqbal666:
that's understandable.
That's what i was thinking, but we can tackle the higher issues first, point me to it. :)
<img width="928" alt="image" src="https://user-images.githubusercontent.com/30576574/163892533-f4651de3-4cd9-4546-a78f-16b26a66a82e.png">

dasyad00:
Hi, I'm also interested in helping out with this PR.

Is the vision mainly to have media controls? Do we want the user to only start the Quran playback from the phone or also the car infotainment? I feel the latter will require a lot of UI/UX considerations especially to reduce distractions for the driver.

benomaire:
[...]

PR #2940, opened 2024-11-10
Fix Android Automotive build

ahmedre:


jibraniqbal666:
Salam @ahmedre, Sorry I have been out again.
What is left to do? I have now lost the context, I would be great if we could start a discussion on github and update the progress there. :)

ahmedre:
wa3laikum alsalam - updated in #1086.

PR #2977, opened 2024-11-30
Move analytics to Gradle scripts and minor cleanup

ahmedre:


ahmedre:
`:app:dependencies` runs with `--configuration madaniReleaseRuntimeClasspath`, which is why this now shows Firebase dependencies, since we don't pass the `disableFirebase` or the `disableCrashlytics`.

PR #3025, opened 2024-12-16
Fix layout direction of the toolbar in RTL

ahmedre:


ahmedre:
cc @khaledkhaled0909

Screenshot_20241217_001058

PR #3036, opened 2024-12-28
Fix audio settings button missing on some devices

ahmedre:
This patch fixes the audio settings button going missing on some devices
(devices with width of 320dp, for example). It sets a width per button
to ensure that everything fits, and assigns weight of 1 to each so that
the buttons can have more space if the space available is more.


ahmedre:
before (Nexus S emulator in portrait)
no_settings_bug

after (Nexus S emulator in portrait)
Screenshot_20241228_234253

after (Nexus S emulator in landscape)
Screenshot_20241228_234347

after (Pixel 9 Pro emulator in landscape)
Screenshot_20241228_234219

Issue #3122, opened 2025-03-25
Bug: Arabic language is reset on rotation and activity change


Raviu56:
When  Arabic language is selected as the app language, it's is reset on rotation and activity change if the Android language settings is not Arabic -  English for example, the app language is set to language until it's closed and open again.

hithamalbasheir:
Just checked your issue, and managed to reproduce it on Android 11

Just to make sure we're on the same page, this is your issue, right?

[https://imgur.com/a/TWaU4qC](Bug reproduce video)

hithamalbasheir:
I'll be picking this up @ahmedre

ahmedre:
one thing that worked well for me in another app was:

```kotlin
  override fun attachBaseContext(newBase: Context) {
    val configuration = Configuration().apply {
      // workaround platform bug pre api 26
      fontScale = 0f
      setLocale(LocaleUtils.getLocale())
    }
    super.attachBaseContext(newBase.createConfigurationContext(configuration))
  }
```

perhaps consider that in the `PagerActivity` class.

hithamalbasheir:
Yep, I think they're related
Thanks for the suggestion, Ahmed!
I'll start working on it and keep you updated, @Raviu56

PR #3190, opened 2025-05-25
Implement app specific language preferences

### LINKED ISSUE #3187: Arabic numbers are showing instead of indian when the language is set to Arabic after upgrading to Android 16 QPR1
Between aya it should be showing indian numbers, even the page number, at least that was the case for me before I updated to Android 16.

Note that the app is set to Arabic for me.

Image

ahmedre:
This also fixes #3187 by switching the locale from simply "ar" to also
passing in a country code, which fixes numbers on Android 16 betas.


ahmedre:
This should also fix #3122.

PR #3201, opened 2025-06-01
Support day mode on the Quran index screen

ahmedre:


ahmedre:
Screenshot_20250601_201817
Screenshot_20250601_201825

ahmedre:
still need some work on some of the dialogs (jump, tag edit) and then for the other screens, but generally working al7amdulillah. additional work will go into future PRs in sha' Allah.

PR #3206, opened 2025-06-03
Improve edge to edge for QuranActivity

ahmedre:
Make the list scroll underneath the bottom bar and pad at the bottom.


OsakaRider:
Thank you for making the nav bar transparent, more devs need to focus on this

PR #3224, opened 2025-06-07
Improvements and fixes to bookmarks code

ahmedre:
Don't construct the bookmark rows in the presenter since the context is
incorrect (application context instead of an activity context). Instead,
let the fragment get the ready raw data back and map it to the
corresponding bookmark rows.


ahmedre:
two notes here:
1. this isn't an ideal solution at all, but I also didn't want to invest an insane amount of time into this today since this should be rewritten "properly" (not just converted to Kotlin, should probably a Kotlin/Compose implementation, and should this idea of sharing unrelated adapter across 3 fragments just doesn't seem correct in retrospect, but hey, the class has been around for more than 10 years now, so it's ok 😂)
2. I got Claude Code to do some bits of this. it made a bunch of bugs in the predict method and walked it through some fixes in various additional prompts until I got tired and fixed it myself. it was helpful though.

ahmedre:
btw, the whole reason this patch was done was that on some api versions, ex 27, changing the language would let the bookmarks still use the old language. this is because the application context wasn't updated with the new context for the app and so the data remained cached.

PR #3240, opened 2025-06-15
Use ExoPlayer instead of MediaPlayer

ahmedre:
ExoPlayer is much more resilient and customizable and is the recommended
route today. For Quran specifically, it has improved seeking, which
stops vbr audio seeks from being extremely inaccurate, and stops us from
having to force encode everything as cbr.

It also handles things for us (like audio focus and ducking) resulting
in less code to maintain. Since it is in use for Android Auto also,
there is no additional apk size to pay on top of today's apk size.


ahmedre:
Thanks to Claude Code for pairing with me on this one. Got it to start working while thoroughly reviewing the changes and asking it questions and redirecting it along the way. Did manual review and changes, and found and fixed one pretty critical bug, but otherwise all good al7amdulillah.

This was something I wanted to do for quite a while now (I have a draft branch with a proof of concept on this since December 26, 2019). The motivation to finally do it now was that some of the new Qaris have misaligned timings, which, after investigation, made me realize the times are correct but the seeking is incorrect (i.e. on a desktop with an audio player, seeking to the same spot results in better results).

I had two options - migrate these files to be cbr or try ExoPlayer which is better about seeking, so went with the latter. Tested the same scenario and al7amdulillah it's working much better.

app/src/main/java/com/quran/labs/androidquran/service/AudioService.kt
```diff
@@ -1,63 +1,37 @@
-/*
```
ahmedre, line 1: this was the case a very long time ago, today, the code likely looks very different.

nacer80:
JazakumAllah khairan

sayyid97:
Mashaalloh, Allohumma baarik 👍

Hamza1244:
@ahmedre Please answer me 4 very good blood Quran I do not know how to send you to add

PR #3304, opened 2025-08-17
Let Renovate update build JDK

ahmedre:
Also migrate to json5 Renovate configuration, and move it into .github.


ahmedre:
https://jakewharton.com/using-renovate-to-update-build-jdk/

PR #3314, opened 2025-08-24
Fix gapped qaris not playing

### LINKED ISSUE #3297: Gapped qari no download no play
The new source code of the Quran Android does not support the (gapped qari) added without a database? 

Like this video:
https://github.com/quran/quran_android/issues/3085#issue-2851979166

ahmedre:
Fixes #3297.


ahmedre:
note that there's a ui bug where loading stays on the audio bar without switching away when switching between files (both in gapped and gapless). since that bug is related to the ExoPlayer migration and affects both gapped and gapless, will fix it in a separate PR in sha' Allah.

PR #3347, opened 2025-09-12
Allow toolbar and audio bar height to change

### LINKED ISSUE #3230: Page number cut off
s23 ultra, s24 ultra, s25 ultra and others:
Image

so far many reports are using the old madani images. I don't think these are issues about "the bar doesn't disappear when tapping," it's more of the text is cut off. got a video from someone with s24 ultra and others, bars disappear when tapping but the text is cut off like in the screenshot.

https://github.com/user-attachments/assets/dd09d0c5-0485-42a3-9211-dc7a7caf0e32

ahmedre:
Instead of forcing the toolbar and audio bar to be the default, make
that the minimum height and wrap content to support cases where font
sizes or display sizes are increased.

Fixes #3230.

### DIFF
--- a/app/src/main/res/layout/quran_page_activity.xml
+++ b/app/src/main/res/layout/quran_page_activity.xml
@@ -29,7 +29,8 @@
     <androidx.appcompat.widget.Toolbar
         android:id="@+id/toolbar"
         android:layout_width="match_parent"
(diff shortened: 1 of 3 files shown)


ahmedre:
Screenshot_20250913_003230
Screenshot_20250913_003250
Screenshot_20250913_003535

PR #3348, opened 2025-09-12
Show sura numbers in search results

ahmedre:
Implement a similar feature request to quran/quran-ios#711 where the
search results show the sura number and not just its name.


ahmedre:
Screenshot_20250913_020251

PR #3351, opened 2025-09-13
Downgrade Firebase to support minSdk 21

ahmedre:
Temporarily downgrade Firebase to support minSdk 21. In sha' Allah we'll
bump again and increase minSdk to 23 after making a release first.


ahmedre:
cc @amirian

PR #3355, opened 2025-09-14
Enable optimized resource shrinking

ahmedre:
This saves around ~0.7mb on a madani release apk, which is nice!


ahmedre:
see article [here](https://android-developers.googleblog.com/2025/09/improve-app-performance-with-optimized-resource-shrinking.html).

PR #3396, opened 2025-10-14
fix: Make QariList BottomSheetScaffold container color transparent

ahmed-okkio:
After the dependency upgrade here #3388, the Qari list started leaving the container background when closed. This change fixed it for me but I'm unsure why this happened so this might not be the best solution

image

image


ahmedre:
JazakumAllah khairan - saw this in another project also and fixed it the same way

PR #3414, opened 2025-10-23
Remove lock portrait/landscape flags

ahmedre:
Android 16 ignores requested orientation calls on larger screens, and
this will likely extend to other devices in the future. The reader can
continue setting these using Android settings for the overall device
orientation (and some devices likely have per-app specific orientation
settings as well).


ahmedre:
see [here](https://android-developers.googleblog.com/2025/01/orientation-and-resizability-changes-in-android-16.html) for more details.

PR #3513, opened 2026-02-04
feat(audio): migrate AudioService to MediaLibraryService

AhmedGamal92:
Migrates the `AudioService` from the legacy `Service` to `androidx.media3.session.MediaLibraryService`.

This change introduces `MediaLibrarySession` to handle media playback, replacing parts of the older `MediaSessionCompat` implementation. A new `QuranServiceCallback` class is also added for future integration with media controllers like Android Auto.

The motivation behind this migration is to build on top of it to enable a better android auto support (check the issue #1086), the end goal is to use the same audio service to sync the player items (surahs) and playback state with android auto, similar to other media playback apps.

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/service/AudioService.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/service/AudioService.kt
@@ -2,7 +2,6 @@ package com.quran.labs.androidquran.service
 
 import android.app.NotificationManager
 import android.app.PendingIntent
-import android.app.Service
 import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
@@ -13,7 +12,6 @@ import android.net.Uri
 import android.os.Build
 import android.os.Handler
 import android.os.HandlerThread
-import android.os.IBinder
 import android.os.Looper
 import android.os.Message
 import android.os.Process
@@ -52,6 +50,8 @@ import androidx.media3.exoplayer.source.DefaultMediaSourceFactory
 import androidx.media3.exoplayer.metadata.MetadataOutput
 import androidx.media3.exoplayer.text.TextOutput
 import androidx.media3.exoplayer.video.VideoRendererEventListener
+import androidx.media3.session.MediaLibraryService
+import androidx.media3.session.MediaSession
 import com.quran.data.core.QuranInfo
 import com.quran.labs.androidquran.QuranApplication
 import com.quran.labs.androidquran.R
@@ -93,7 +93,9 @@ import androidx.core.net.toUri
  * (which come from our main activity, [PagerActivity], which signal
  * the service to perform specific operations: Play, Pause, Rewind, Skip, etc.
  */
-class AudioService : Service(), Player.Listener {
+class AudioService : MediaLibraryService(), Player.Listener {
+
+  private var mediaLibrarySession: MediaLibrarySession? = null
 
   // our exo player
   private var player: ExoPlayer? = null
@@ -154,6 +156,7 @@ class AudioService : Service(), Player.Listener {
   private var currentWord: Int? = null
   private val compositeDisposable = CompositeDisposable()
   private lateinit var scope: CoroutineScope
+  private val quranServiceCallback = QuranServiceCallback()
 
   @Inject
   lateinit var quranInfo: QuranInfo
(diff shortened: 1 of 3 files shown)


AhmedGamal92:
@ahmedre 
SA
I'm not sure what is the usual practice you follow when such migration is needed, two questions I have:
- While I did test the audio service functionality on couple of devices and things are working the same, how do we ensure proper QA to avoid any side effects ?
- As this PR by itself doesn't introduce extra functionalities, I see it better to merge and test that migration individually before introducing the later changes specific to Android Auto, as also this one affects the main app, wdyt ?

ahmedre:
wa3laikum alsalam - jazakumAllah khairan for the pr!
will read this during the weekend in more detail in sha' Allah. to answer your question though, agree with the second point. unfortunately for the first point, don't have any tests or anything and it's a bit difficult to test some of this stuff (integration with the service, etc) so don't have any great ideas there.

app/src/main/java/com/quran/labs/androidquran/service/AudioService.kt
```diff
@@ -261,8 +265,20 @@ class AudioService : Service(), Player.Listener {
...
+      // Initialize ExoPlayer
+      val player = makeOrResetExoPlayer()
+
+      // Initialize MediaLibrarySession
+      mediaLibrarySession = MediaLibrarySession.Builder(this, player, quranServiceCallback)
+        .build()
+    }
   }
 
+  override fun onGetSession(controllerInfo: MediaSession.ControllerInfo): MediaLibrarySession? = mediaLibrarySession
```
AhmedGamal92: Found an issue here that we will face later, as we're creating the session async within the handler, it happens that the session can be null if getSession is called directly after onCreate, this is the case when a media controller is created and the service is not running already (eg. when we have android auto synced, when the app is opened from auto and was not active before)

With the current approach, it has to happpen with the handler thread as that's where ExoPlayer will be used from in all other operations

Quite a tricky issue, a proper solution IMO would be to get rid of the handler and the messaging posting mechanism we have in the service, the  `MediaLibraryService` is already intended for background play we don't need a bg thread, check the example [here ](https://developer.android.com/media/implement/playback-app) also using ExoPlayer.

@ahmedre wdyt before trying out that direction ?

ahmedre:
looks good to me, جزاكم الله خيراً

minor points:
1. as you suggested, fixing the potential null return from `onGetSession`
2. we override `onStartCommand` and return `START_NOT_STICKY`, when it seems like the underlying `MediaSessionService` always returns `START_STICKY`. 

can you try to test media buttons to make sure they still work (if you have headphones that can pause/resume playback etc)?

app/src/main/java/com/quran/labs/androidquran/service/AudioService.kt
AhmedGamal92: Ok cool, I'll then get rid of the handler and update the PR.

AhmedGamal92:
> looks good to me, جزاكم الله خيراً
> 
> minor points:
> 
> 1. as you suggested, fixing the potential null return from `onGetSession`
> 2. we override `onStartCommand` and return `START_NOT_STICKY`, when it seems like the underlying `MediaSessionService` always returns `START_STICKY`.
> 
> can you try to test media buttons to make sure they still work (if you have headphones that can pause/resume playback etc)?

@ahmedre pushed a commit getting rid of the serviceHandler as discussed earlier, please check now

ahmedre:
wa3laikum alsalam wa ra7matullah,
jazakumAllah khairan!
2 things:
1. I think we'll be double handling media buttons this way (via `MediaButtonReceiver.handleIntent`), and I think `MediaLibraryService` already handles this for us.
2. secondly, in our `relaxResources` and `makeOrResetExoPlayer`, we never re-bind the session to the new player using `mediaLibrarySession?.setPlayer(`.

AhmedGamal92:
> wa3laikum alsalam wa ra7matullah, jazakumAllah khairan! 2 things:
> 
> 1. I think we'll be double handling media buttons this way (via `MediaButtonReceiver.handleIntent`), and I think `MediaLibraryService` already handles this for us.
> 2. secondly, in our `relaxResources` and `makeOrResetExoPlayer`, we never re-bind the session to the new player using `mediaLibrarySession?.setPlayer(`.

True, good catch, we should cleanup the mediaSessionCompat, the MediaLibraryService should handle that for us, updating the PR soon.

AhmedGamal92:
@ahmedre 
SA, it's been awhile, I did update the PR now getting rid of the legacy media session, check now and let me know.
I did some testing and working fine with me.

PR #3530, opened 2026-02-15
Improve behavior of double page mode on foldables

### LINKED ISSUE #3526: Double page bug on foldables
I'm using a pixel 10 pro fold. In the unfolded mode, i can read 2 pages at once, but as soon as I slightly fold the phone it goes to single page which is quite bothersome. I would expect the double page mode to stay the same until the phone is completely folded. I usually use the phone slightly bent so i can hold it in 1 hand while reading the Quran.

The single front screen shows 1 page accurately.

It seems like the transition from dual page to single page on foldables is happening too early when closing the phone.

ahmedre:
When a foldable is partially folded, the app would switch to single page
mode. The guidance from Google for handling "book posture" is to handle
both half opened and flat states, so that dual page mode can remain
until the device is fully closed.

Fixes #3526.


ahmedre:
see the documentation [here](https://developer.android.com/develop/ui/compose/layouts/adaptive/foldables/make-your-app-fold-aware#book_posture).

PR #3534, opened 2026-02-15
Add album art to Android Auto

ahmedre:
Refs #1086.


ahmedre:
CleanShot 2026-02-16 at 03 40 30@2x
CleanShot 2026-02-16 at 03 40 56@2x
CleanShot 2026-02-16 at 03 41 04@2x
CleanShot 2026-02-16 at 03 41 09@2x
Screenshot_20260216_034659

ahmedre:
holding off until I replace this implementation with something more efficient - this works but is very expensive both in terms of cpu and memory (since we effectively block until we generate every image). have a partially working solution to that so will update before merging in sha' Allah.

alti3:
> holding off until I replace this implementation with something more efficient - this works but is very expensive both in terms of cpu and memory (since we effectively block until we generate every image). have a partially working solution to that so will update before merging in sha' Allah.

why not vectorize the images to SVGs.
114 SVGs, one for each surah
how large can it be?

ahmedre:
That'd also work but the implementation here is good also after the updates so we can merge it

PR #3547, opened 2026-02-23
Add reading history feature with date, time and page navigation

Am01bathawab:
This adds a reading history screen that shows all pages 
visited with date and time. Tapping any history entry 
navigates directly to that page. 

Changes made:
- Added new reading_history database table
- Added ReadingHistory.sq with queries
- Updated BookmarksDBAdapter with history functions
- Updated RecentPageModel to expose history
- Created ReadingHistoryActivity to display history
- Registered activity in AndroidManifest
- Updated RecentPagePresenter to open history screen

The history is accessible via the book icon in the 
reading screen.


ahmedre:
we already have this in the app from long ago - see [LastPage.sq](common/bookmark/src/main/sqldelight/com/quran/mobile/bookmark/LastPage.sq) - and we keep the date and time for the last 3 entries today.

please open an issue before implementing new features in the future so you do not do work that is already done or has a plan associated with it.

جزاكم الله خيراً for the pr

PR #3552, opened 2026-02-24
Revert "Remove lock portrait/landscape flags"

ahmedre:
This reverts commit 2f2dda5d9912be1a1622e3c46eb68b1afa47d1f1.


ahmedre:
Too many emails with people complaining so bringing this back. Will open another patch hiding this on Android 16+ because it won't work there - see [here](https://android-developers.googleblog.com/2025/01/orientation-and-resizability-changes-in-android-16.html) for details.

this, combined with the fact that phones have this setting built in anyway, make this unnecessary in the app, but  ¯\_(ツ)_/¯

PR #3734, opened 2026-07-10
Add pause-between-verses playback option

NourEddineX:
## Summary

Adds a configurable pause duration between ayahs during audio playback. The motivation is to give reciters time to repeat each verse aloud after it plays — useful for memorization and recitation practice.

- New pause duration option in the audio panel's playback settings (`AyahPlaybackFragment`), plumbed through `AudioRequest` into `AudioService`.
- `AudioService` honors the configured gap between verses during gapped playback.
- Fixes a media-session flicker where the inter-verse gap briefly reported a Paused state to the media session.

## Test plan

- Play a sura with pause-between-verses set to various durations; verify the gap between ayahs matches.
- Verify default (no pause) behavior is unchanged.
- Verify media session / notification no longer flickers Paused during the inter-verse gap.

https://github.com/user-attachments/assets/ace01fa8-a5f7-420c-9f40-d50003ab150e

### DIFF
--- a/app/src/main/java/com/quran/labs/androidquran/feature/reading/presenter/AudioPresenter.kt
+++ b/app/src/main/java/com/quran/labs/androidquran/feature/reading/presenter/AudioPresenter.kt
@@ -40,7 +40,8 @@ constructor(
     rangeRepeat: Int,
     enforceRange: Boolean,
     playbackSpeed: Float,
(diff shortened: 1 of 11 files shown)


NourEddineX:
@ahmedre Please review, JAK

PR #3438, opened 2025-11-17
feat(page): add customizable page background selection system

aalhendi:
Fixes #2737 
Fixes #648 
Fixes #647 
Fixed #3491 
Ref https://github.com/quran/quran_android/issues/2211#issuecomment-1454561374

# Summary

Replaces the gradient background system with a theme selection bottom sheet, matching the design pattern from the iOS sister app.

# Features

**6 Page Themes:**
- **Paper** - Cream background
- **Original** - Pure white
- **Calm** - Parchment
- **Focus** - Off-white 
- **Quiet** - Dark grey with white text
- **Black** - Pure black with white text

**App Theme Dropdown:**
- Light / Dark / System default
- Controls overall app UI theming

**Dark Theme Support:**
- Quiet and Black themes automatically invert Quran page text to white
- Uses `ColorMatrixColorFilter` (same as existing night mode)

## UI
- Bottom sheet with 3-column grid of theme cards
- Each card shows Arabic preview text "نور" with theme colors
- Selection border indicates active theme
- App Theme dropdown above the grid

# Screenshots
image

PR #3764, opened 2026-08-07
Add 'Green' option to background color preferences

mohamedelghazali:
Added a new background color option 'Green' for user preferences.


ahmedre:
we want to go with something like #3438 for consistency with web and iOS instead of doing this specifically.

mohamedelghazali:
We can work on that

PR #3765, opened 2026-08-08
Clean up strings.xml by removing comments

mohamedelghazali:
Removed commented sections for clarity and maintenance.


ahmedre:
comments aren't really bothering anyone - not against removing them but don't really see a point either, they're compiled out  ¯\_(ツ)_/¯

PR #3804, opened 2026-09-02
fix: correct French sura name

mohamede1945:
## Summary
- correct the French name for Al-Furqan from TLe discernement to Le discernement

## Validation
- xmllint --noout common/ui/core/src/main/res/values-fr/sura_names.xml
- git diff --check


ahmedre:
جزاكم الله خيراً

PR #3826, opened 2026-09-11
#1407 improvements to audio settings

DoozyDoz:
supersedes #1429 (from my old account)

fixing the two things from #1407, with the feedback from #1429 addressed:

1. when changing audio settings while playing, audio restarts from the beginning.
- in AudioService, check if the current playing ayah is still within the new range, and if so keep playing without restarting. if it's out of range or start ayah changed, it stops and starts from the new start.

2. remember "play only above verses" setting instead of having to manually check it every time.
- saved the setting in shared preferences through QuranSettings when apply is clicked, and restore it in refreshView in AyahPlaybackFragment. no gson or extra dependencies.


DoozyDoz:
Dear @ahmedre Kindly find time and review

